PR #1, opened 2021-04-16
initial commit

maltesander:


crd/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+use stackable_operator::Crd;
+
+#[derive(
+    Clone, CustomResource, Debug, Default, Deserialize, Eq, Hash, JsonSchema, PartialEq, Serialize,
+)]
+#[kube(
+    group = "opa.stackable.tech",
+    version = "v1",
+    kind = "OpaServer",
+    shortname = "os",
```
soenkeliebau, line 13: nit: wouldn't "opa" be the obvious choice here?

server/packaging/rpm/SOURCES/stackable-opa-operator-server-VERSION/usr/lib/systemd/system/stackable-opa-operator-server.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Stackable OpenPolicyAgent Operator
+Before=
+After=network.target
+[Service]
+User=root
+ExecStart=/opt/stackable/opa-operator/opa-operator-server
```
soenkeliebau, line 7: The binary here needs to match the parameter you pass to the buildrpm script in the nightly action.

server/packaging/debian/service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Stackable OpenPolicyAgent Operator
+Before=
+After=network.target
+[Service]
+User=root
+ExecStart=/opt/stackable/opa-operator/opa-operator-server
```
soenkeliebau, line 7: See comment in other service file about the binary name.

soenkeliebau (changes requested):
Left two comments (actually the same comment, but in two places) :)

crd/src/lib.rs
maltesander: Yeah, any changes in the group name then? 
"opa.opa.stackable.tech" is not my favorite :)

server/packaging/rpm/SOURCES/stackable-opa-operator-server-VERSION/usr/lib/systemd/system/stackable-opa-operator-server.service
maltesander: Thanks will change.

soenkeliebau (approved):
LGTM

PR #5, opened 2021-05-02
Removed hard coded label selector schema and used the one provided in…

maltesander:
… operator rs. Added binary to generate OPA CRD.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1370,7 +1370,6 @@ dependencies = [
  "schemars",
  "serde",
  "serde_json",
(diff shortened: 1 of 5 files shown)


lfrancke (approved):
I'm only on my phone and it's hard to check: Can you also remove some dependencies from cargo. Toml? Like serde-json? Or is it used elsewhere

maltesander:
I removed some dependencies. serde_json is required both in operator and crd.

lfrancke (approved):

PR #18, opened 2021-05-17
Add opa reference for other operators

maltesander:


crd/src/util.rs
```diff
@@ -0,0 +1,674 @@
...
+                return Err(PodWithoutHostname {
+                    pod: pod_name.clone(),
+                });
+            }
+            Some(node_name) => node_name,
+        };
+
+        let role_group = match pod
+            .metadata
+            .clone()
```
lfrancke, line 275: The clone is dirty but I see why you're doing it.
Removing it would mean a nested match I believe.
Up to you. Fine for me as it is.

crd/src/util.rs
```diff
@@ -0,0 +1,674 @@
...
+        server_and_port_list.push((node_name, opa_port));
+    }
+
+    // Sort list by hostname to make resulting connection strings predictable
+    // Shouldn't matter for connectivity but makes testing easier and avoids unnecessary
+    // changes to the infrastructure
+    server_and_port_list.sort_by(|(host1, _), (host2, _)| host1.cmp(host2));
+
+    // TODO: randomly search for node? -> for now we just take the last one
+    let server = server_and_port_list.pop();
```
lfrancke, line 312: Yeah, this is not great for load balancing purposes.
Could you pick a random one?

https://stackoverflow.com/questions/34215280/how-can-i-randomly-select-one-element-from-a-vector-or-array/34215930

seems relatively simple

lfrancke (changes requested):

crd/src/util.rs
```diff
@@ -0,0 +1,674 @@
...
+//    for completeness and may not be exhaustive with regard to configuration.
+#[derive(Clone, Debug, Deserialize, JsonSchema, Serialize)]
+pub enum OpaApi {
+    /// e.g. GET /v1/policies/<id>
+    Policy { id: String },
+    /// e.g. GET /v1/data/<package_path>/<rule>
+    Data {
+        /// e.g. /kafka/authz
+        package_path: String,
+        /// e.g. "/allow"
```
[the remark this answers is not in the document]
lfrancke: Either I'm missing something or you didn't actually change anything here?

crd/Cargo.toml
```diff
@@ -24,7 +25,7 @@ url = "2"
 [dev-dependencies]
 k8s-openapi = { version = "0.11", default-features = false, features = ["v1_20"] }
 indoc = "1.0"
-rstest = "0.8"
+rstest = "0.9"
```
lfrancke, line 28: In the meantime... 0.10 has been released :)

crd/src/util.rs
```diff
@@ -174,10 +174,28 @@ pub async fn get_opa_connection_info(
     Ok(OpaConnectionInformation { connection_string })
 }
 
+/// Remove redundant "/" from the url path
```
lfrancke, line 177: What does "redundant" mean here?
I don't understand the code to be honest.

crd/src/util.rs
```diff
@@ -308,8 +318,8 @@ fn get_opa_connection_string_from_pods(
     // changes to the infrastructure
     server_and_port_list.sort_by(|(host1, _), (host2, _)| host1.cmp(host2));
 
-    // TODO: randomly search for node? -> for now we just take the last one
-    let server = server_and_port_list.pop();
+    let index = rand::random::<usize>() % server_and_port_list.len();
+    let server = server_and_port_list.get(index);
```
lfrancke, line 322: There's a method for this: https://stackoverflow.com/questions/34215280/how-can-i-randomly-select-one-element-from-a-vector-or-array/34215930

crd/src/util.rs
```diff
@@ -508,15 +530,15 @@ mod tests {
...
+    Box::new(["https://", "ebian:8181/v1/data/some_package/some_sub_package/allow"])
     )]
     fn get_connection_string(
         #[case] opa_spec: &str,
         #[case] opa_pods: &str,
         #[case] opa_api: &OpaApi,
         #[case] opa_api_protocol: &OpaApiProtocol,
         #[case] desired_node_name: Option<String>,
-        #[case] expected_result: &str,
+        #[case] expected_result: Box<[&str]>,
```
lfrancke, line 541: Why all the boxes?

lfrancke (changes requested):

crd/src/util.rs
maltesander: Thats for the created opa url. I hardcoded the slashes in the format!(...) and now you can e.g. refer to the rule either like:
/package/ or package/ etc.

crd/src/util.rs
maltesander: For the 4. test case i need to check if it contains https and the correct opa url (because of the introduced randomness). Therefore i need an array and could not make it work unboxed (not sized etc)

crd/src/util.rs
lfrancke: Okay, could you document this?

Something like:
Removes slashes ("/") from the beginning or end of a string

But...I don't really understand the code so it's up to you to describe what the function actually does :)

crd/src/util.rs
lfrancke: Ah! In that case it makes more sense to take a slice of string slices:

```suggestion
        #[case] expected_result: &[&str],
```

PR #29, opened 2021-06-01
Add Antora docs

lfrancke:


docs/antora.yml
```diff
@@ -0,0 +1,5 @@
+name: opa
+version: master
```
maltesander, line 2: If it refers to the branch, it should be "main". I checked some other operators and it is "master" everywhere, so if im wrong just ignore this.

README.adoc
```diff
@@ -1,10 +1,7 @@
-= Stackable Open Policy Agent Operator
+= Stackable Operator for OpenPolicyAgent
 
-This Kubernetes Operator deploys and manages the Open Policy Agent (OPA).
+This is a Kubernetes Operator to manage OpenPolicyAgent servers.
```
maltesander, line 3: I think we should use the shortcut OPA somewhere on top. I think many people can relate better to the shortcut (at least in my case).

maltesander (approved):
Just two small comments, you can ignore them, so approved.

docs/antora.yml
lfrancke: Yeah, `master` is not the branch but an Antora specific thing.

maltesander: Ok did not know that ;)

README.adoc
lfrancke: I've changed it

maltesander (approved):

PR #37, opened 2021-07-08
Adapt to operator-rs, kube 0.57 and product config

maltesander:
This PR contains multiple changes:

- Adapt to latest operator-rs and kube 0.57
- Add product config validation and properties file
- Add graceful wait if CRD is missing
- Changed version "v1" to "v1alpha1"


operator/src/pod_utils.rs
```diff
@@ -0,0 +1,40 @@
+/// All pod names follow a simple pattern: spark-<cluster_name>-<role_group>-<node_type>-<node_name>
```
lfrancke, line 1: Yeah...this will need to change...as discussed. But not now.

operator/src/lib.rs
```diff
@@ -235,36 +162,152 @@ impl OpaState {
...
+                        if let Some(repo_reference) = config.get(REPO_RULE_REFERENCE) {
+                            cm_data.insert(
+                                CONFIG_FILE.to_string(),
+                                create_config_file(repo_reference),
+                            );
+                            config_maps.push(
+                                ConfigMapBuilder::new()
+                                    .metadata(
+                                        ObjectMetaBuilder::new()
+                                            .name(cm_name.clone())
```
lfrancke, line 234:
```suggestion
                                            .name(&cm_name)
```

operator/src/lib.rs
```diff
@@ -235,36 +162,152 @@ impl OpaState {
...
+                            self.create_config_map(config_map).await?;
+                        }
                     }
                 }
             }
         }
         Ok(ReconcileFunctionAction::Continue)
     }
+
+    async fn create_pod_and_config_maps(
```
lfrancke, line 200: Could you add a comment on the logic in here?
Something that explains what's going on, how many configmaps there will be etc.
I know this code is obvious to you now but it's a lot of code without a single comment

operator/src/lib.rs
```diff
@@ -235,36 +162,152 @@ impl OpaState {
...
+                                            .name(cm_name.clone())
+                                            .ownerreference_from_resource(
+                                                &self.context.resource,
+                                                Some(true),
+                                                Some(true),
+                                            )?
+                                            .namespace(&self.context.client.default_namespace)
+                                            .build()?,
+                                    )
+                                    .data(cm_data.clone())
```
lfrancke, line 243: Is this correct?
I guess it is. But I believe you should move the initialization of cm_data in this `if` branch and then you can move it in as well (no need to clone). It's not used elsewhere, is it?

operator/src/lib.rs
```diff
@@ -235,36 +162,152 @@ impl OpaState {
...
+            &self.context.resource.spec.version.to_string()
+        ));
+        container_builder.command(start_command);
+        container_builder.add_configmapvolume(cm_name, "conf".to_string());
+
+        for env in env_vars {
+            if let Some(val) = env.value {
+                container_builder.add_env_var(env.name, val);
+            }
+        }
```
lfrancke, line 293: The builder should have a method for this

lfrancke (changes requested):
Minor comments

operator/src/lib.rs
maltesander: Yeah i noted some stuff that were missing for me in the builders.

crd/src/lib.rs
```diff
@@ -49,30 +54,73 @@ pub enum OpaVersion {
...
     pub repo_rule_reference: String,
 }
 
-impl Crd for OpenPolicyAgent {
-    const RESOURCE_NAME: &'static str = "openpolicyagents.authz.stackable.tech";
-    const CRD_DEFINITION: &'static str = include_str!("../../deploy/crd/openpolicyagent.crd.yaml");
+impl Configuration for OpaConfig {
+    type Configurable = OpenPolicyAgent;
+
+    fn compute_env(
```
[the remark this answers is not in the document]
maltesander: Yeah would reduce the amount of unused code.

server/src/bin/generate_crds.rs
```diff
@@ -1,7 +0,0 @@
-use stackable_opa_crd::OpenPolicyAgent;
-use std::error::Error;
-
-fn main() -> Result<(), Box<dyn Error>> {
-    println!("{}", serde_yaml::to_string(&OpenPolicyAgent::crd())?);
```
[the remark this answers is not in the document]
maltesander: There is only one in dev-dependencies which is required.

operator/src/lib.rs
maltesander: Its used in line 287:
`container_builder.add_configmapvolume(cm_name, "conf".to_string());`

operator/src/lib.rs
maltesander: Yeah now where i see it, its not optimal. The thought was if you define multiple files (that go into one config map) we need to define data outside the loop, fill it inside and build the config map outside the loop too. Ill try to adapt.

lfrancke (approved):

razvan (approved):
ok

PR #50, opened 2021-08-02
Fix clippy warnings in 1.54

lfrancke:
## Description

Rust 1.54 introduced some new Clippy warnings. This PR fixes all of those


maltesander (approved):
LGTM.

PR #74, opened 2021-09-10
Update templated files to rev 57d7d55

soenkeliebau:
Automatically created PR based on commit 57d7d559a519277a08d5a7453b59b227c3638992 in stackabletech/operator-templating repo.

Original commit message:
Moved repository and retired file definitions to separate file.


soenkeliebau:
Still errors in the Dockerfile to solve, but getting there...

PR #146, opened 2021-12-07
STFU migration

teozkr:
## Review Checklist
- [x] (Integration-)Test cases added (or not applicable)
- [x] Documentation added (or not applicable)
- [x] Changelog updated (or not applicable)


rust/operator/src/discovery.rs
```diff
@@ -0,0 +1,70 @@
...
+                    opa_version(opa).as_deref().unwrap_or("unknown"),
+                    &OpaRole::Server.to_string(),
+                    "discovery",
+                )
+                .build(),
+        )
+        .add_data("OPA", url)
+        .build()
+        .unwrap())
+}
```
maltesander, line 70: Catch the error properly?

rust/operator/src/discovery.rs
```diff
@@ -0,0 +1,70 @@
...
+                    opa: ObjectRef::from_obj(opa),
+                })?
+                .with_recommended_labels(
+                    opa,
+                    APP_NAME,
+                    opa_version(opa).as_deref().unwrap_or("unknown"),
+                    &OpaRole::Server.to_string(),
+                    "discovery",
+                )
+                .build(),
```
maltesander, line 65: (We discussed already,) but i still dont like only the service here.

Would it make sense to additionally create configmaps per opa pod and add a "host" label? 
That way we could select the local opa via labels?

Since we have one pod per node anyways, should not result in a back and forth when updating or sth is changing?

maltesander (commented):
Just some zookeeper artifacts and discovery discussion.
Didnt test yet (will do tomorrow).

rust/operator/src/discovery.rs
teozkr: Good point. This was fixed in ZK but that fix needs to be carried over into here.

rust/operator/src/discovery.rs
teozkr: > Would it make sense to additionally create configmaps per opa pod and add a "host" label? That way we could select the local opa via labels?

How would the pods know which configmap to mount in? Pods can only mount named `ConfigMap`s, a search query isn't enough.

If you want to enforce local traffic then you'll either need to use `Service.spec.internalTrafficPolicy` or bake K8s awareness into the containers themselves.

`internalTrafficPolicy` is a pretty heavy hammer (as we discussed before, ideally we'd fan the traffic out onto the cluster if the local pod is down), but that's not really a worse position to be in than the current state of always connecting to `localhost`.

>

maltesander: Yeah youre right, ill resolve this.

rust/operator/src/discovery.rs
```diff
@@ -0,0 +1,70 @@
...
+        source: stackable_operator::error::Error,
+        opa: ObjectRef<OpenPolicyAgent>,
+    },
+    #[snafu(display("object has no name associated"))]
+    NoName,
+    #[snafu(display("object has no namespace associated"))]
+    NoNamespace,
+}
+
+/// Builds discovery [`ConfigMap`]s for connecting to a [`ZookeeperCluster`] for all expected scenarios
```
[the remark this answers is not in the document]
teozkr: fb684652cd0509c0217cb6cc67bc410839fd1057

rust/operator/src/opa_controller.rs
```diff
@@ -0,0 +1,344 @@
+//! Ensures that `Pod`s are configured and running for each [`ZookeeperCluster`]
```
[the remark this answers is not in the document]
teozkr: fb684652cd0509c0217cb6cc67bc410839fd1057

rust/operator/src/opa_controller.rs
```diff
@@ -0,0 +1,344 @@
...
+    Ok(ReconcilerAction {
+        requeue_after: None,
+    })
+}
+
+/// The server-role service is the primary endpoint that should be used by clients that do not perform internal load balancing,
+/// including targets outside of the cluster.
+///
+/// Note that you should generally *not* hard-code clients to use these services; instead, create a [`ZookeeperZnode`](`stackable_zookeeper_crd::ZookeeperZnode`)
+/// and use the connection string that it gives you.
```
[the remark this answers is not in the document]
teozkr: fb684652cd0509c0217cb6cc67bc410839fd1057

rust/operator/src/discovery.rs
teozkr: c1b8789cddf0e37261d48948a563c22c6a3a592f

maltesander (approved):
LGTM.

PR #156, opened 2021-12-17
Removed configurable port from code and product config.

maltesander:
## Description
- Changed version to string
- Updated dependencies
- Fixed comments
- Removed port from code and product config

## Review Checklist
- [x] (Integration-)Test cases added (or not applicable)
- [x] Documentation added (or not applicable)
- [x] Changelog updated (or not applicable)


deploy/config-spec/properties.yaml
```diff
@@ -6,29 +6,8 @@ spec:
         regex: "^((https?|ftp|file)://)?[-a-zA-Z0-9+&@#}/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]"
         examples:
           - "https://www.stackable.de/blog/"
-    - unit: &unitPort
-        name: "port"
-        regex: "^([0-9]{1,4}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$"
 
 properties:
-  - property: &opaPort
```
teozkr, line 14: Do we want to remove the port from p-c? I thought this was supposed to include the kitchen sink, even if we don't support changing it.

rust/operator/src/opa_controller.rs
```diff
@@ -46,11 +46,11 @@ pub struct Ctx {
...
-    #[snafu(display("object has no namespace"))]
-    ObjectHasNoNamespace,
-    #[snafu(display("object defines no version"))]
-    ObjectHasNoVersion,
-    #[snafu(display("object has no server role"))]
+    #[snafu(display("object {} has no namespace", obj_ref))]
+    ObjectHasNoNamespace { obj_ref: ObjectRef<OpenPolicyAgent> },
+    #[snafu(display("object {} defines no version", obj_ref))]
+    ObjectHasNoVersion { obj_ref: ObjectRef<OpenPolicyAgent> },
+    #[snafu(display("object defines no server role"))]
```
teozkr, line 53: The `obj_ref` is now managed by `controller::Error::ReconcilerFailed` as of kube 0.65.0.

rust/operator/src/opa_controller.rs
```diff
@@ -232,9 +232,10 @@ fn build_server_rolegroup_config_map(
     })
 }
 
-/// The rolegroup [`StatefulSet`] runs the rolegroup, as configured by the administrator.
+/// The rolegroup [`DaemonSet`] runs the rolegroup, as configured by the administrator.
```
teozkr, line 235: Oops, heh

rust/operator/src/opa_controller.rs
```diff
@@ -312,8 +313,13 @@ fn build_server_rolegroup_daemonset(
     })
 }
 
-pub fn opa_version(opa: &OpenPolicyAgent) -> Result<String> {
-    Ok(opa.spec.version.to_string())
+pub fn opa_version(opa: &OpenPolicyAgent) -> Result<&str> {
```
teozkr, line 316: :+1:

deploy/config-spec/properties.yaml
maltesander: Now it was a required property (meaning p-c is complaining if you leave it out). Since we hard code it i think there is no need to have it in there?

rust/operator/src/opa_controller.rs
maltesander: done

teozkr (approved):
LGTM once tests pass

teozkr (approved):

Issue #237, opened 2022-02-17
LDAP user attributes (groups) in OPA


fhennig:
As a user/admin, I want to make policy decisions based on user attributes such as group membership, that I want to define in LDAP. I want to use these user attributes in my RegoRules in OPA.

This is done when
- we have a solution design (ADR?) that solves the issue described above of getting group memberships from LDAP in a scalable way (i.e. not every OPA should contact LDAP every time to get all group information)
- the solution should evaluate push vs. pull as well

(Note: This could be SSSD, but it's only one option)

--- 

**Original below**

A couple of points from our discussion:
- We want the group handling inside OPA to not have to split it across all products, so (1) is out.
- We also do not want to do a lot of LDAP requests constantly so (2c) is also out
- We want to do either (2a) or (2b)

We are still figuring out how we will do that exactly.

---

**Previous issue text for context:**

As a user/admin, I want to make policy decisions based on user attributes such as group membership, that I want to define in LDAP. I want to use these user attributes in my RegoRules in OPA.

OPA already has [documentation](https://www.openpolicyagent.org/docs/v0.11.0/guides-identity/) on how to do this, there are 2 (4) variants:
1. put the user attributes into the policy decision request
2. OPA gets the attributes directly from LDAP
    a. periodically pull from LDAP bundle API
    b. push LDAP into OPA
    c. pull LDAP info on demand

Since not every product might support forwarding LDAP information to OPA, and since we probably have access to the LDAP server anyways, I'm inclined to go with Option 2. Between 2a, 2b and 2c I am unsure. It's worth reading the linked docs for detailed trade-offs, there's a summary table at the end as well outlining some trade-offs.

Some things to consider are:
- The amount of network requests and potential lag when chainging/fowarding/depending on other requests
- Staleness of cached LDAP info
- Size of the bundles being transferred and stored
- Duplication across OPA instances

fhennig:
UserInfoFetcher solves this

PR #433, opened 2023-04-13
User info fetcher

nightkr:
# Description

This injects a new component into the OPAs, which is used to query a directory backend (such as Keycloak) for additional user information, such as groups and roles. See #477. This is a first step towards implementing #237.

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Documentation added or updated
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

Once the review is done, comment `bors r+` (or `bors merge`) to merge. [Further information](https://bors.tech/documentation/getting-started/#reviewing-pull-requests)


nightkr:
This is very much a prototype, before any actual release we'd need to make the backends configurable, and at least have the scaffolding to add support for more directory backends (such as LDAP).

There's also a corresponding Trino integration, over at https://github.com/stackabletech/trino-operator/tree/spike/user-info-fetcher.

fhennig:
heya! Looks like this is related to this long open ticket we have: https://github.com/stackabletech/opa-operator/issues/237

I've got a few questions. Will every OPA instance query the group data on its own? How long is group membership for a user cached?

nightkr:
The current spike doesn't do any caching. We have a few avenues for turning it on eventually, either adding it to the group-fetcher sidecar ourselves, or by turning on OPA's built-in `http.send` cache. The former would give us more control, the latter would be trivial to implement.

Each OPA currently runs independently of each other, so each instance would be run its own independent cache unless we introduce a shared cache component of some kind.

fhennig:
Got it, thanks :ok_hand:

soenkeliebau:
Just so this isn't lost, @Jimvin expressed a keen interest to have ways of invalidating specific caches as well as @lfrancke who I believe mentioned the ability to evict specific users from all caches.
Probably not for version 0, but worth keeping in mind.

Tiltfile
```diff
@@ -22,6 +22,8 @@ watch_file('result')
 if os.path.exists('result'):
    k8s_yaml('result/crds.yaml')
 
+k8s_kind('Deployment', image_json_path='{.spec.template.metadata.annotations.image}')
```
nightkr, line 25: Need to move this into templating...

Tiltfile
```diff
@@ -36,3 +38,6 @@ helm_crds, helm_non_crds = filter_yaml(
    kind = "^CustomResourceDefinition$",
 )
 k8s_yaml(helm_non_crds)
+
+k8s_yaml('docs/modules/opa/examples/getting_started/opa.yaml')
+k8s_resource(new_name='simple-opa', objects=['simple-opa:OpaCluster'], extra_pod_selectors=[{'app.kubernetes.io/name': 'opa'}], port_forwards=['8081:8081', '9476:9476'])
```
nightkr, line 43: Not sure about if this can be templated well, but I found it super convenient.

deploy/helm/opa-operator/templates/deployment.yaml
```diff
@@ -14,11 +14,12 @@ spec:
       {{- include "operator.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-      {{- with .Values.podAnnotations }}
       annotations:
+        image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+        {{- with .Values.podAnnotations }}
```
nightkr, line 19: Move into templating

deploy/helm/opa-operator/templates/deployment.yaml
```diff
@@ -43,6 +44,11 @@ spec:
           env:
             - name: OPA_BUNDLE_BUILDER_CLUSTERROLE
               value: {{ include "operator.fullname" . }}-opa-bundle-builder-clusterrole
+            - name: OPERATOR_IMAGE
+              # Tilt can use recognize annotations as image paths, but not env variables
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.annotations['image']
```
nightkr, line 51: Move into templating

docs/modules/opa/examples/getting_started/opa.yaml
```diff
@@ -4,8 +4,25 @@ kind: OpaCluster
...
 spec:
+  clusterConfig:
+    userInfoFetcher:
+      backend:
+        keycloak:
+          url: http://192.168.122.1:8080
+          credentialsSecretName: opa-keycloak
+          adminRealm: master
+          userRealm: master
+          clientId: admin-cli
```
nightkr, line 15: Getting started should use none backend

nightkr: Not sure if we want to document this yet at all...

docs/modules/opa/examples/getting_started/opa.yaml
```diff
@@ -4,8 +4,25 @@ kind: OpaCluster
...
     roleGroups:
       default: {}
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: opa-keycloak
+stringData:
+  username: admin
+  password: admin
```
nightkr, line 28: Also move out of tutorial

rust/operator-binary/src/controller.rs
```diff
@@ -500,7 +505,11 @@ fn build_server_rolegroup_config_map(
                 ))
                 .build(),
         )
-        .add_data(CONFIG_FILE, build_config_file());
+        .add_data(CONFIG_FILE, build_config_file())
+        .add_data(
+            "user-info-fetcher.json",
+            serde_json::to_string_pretty(&opa.spec.cluster_config.user_info_fetcher).unwrap(),
```
nightkr, line 511: snafu

rust/operator-binary/src/controller.rs
```diff
@@ -533,6 +542,7 @@ fn build_server_rolegroup_daemonset(
     rolegroup_ref: &RoleGroupRef<OpaCluster>,
     server_config: &HashMap<PropertyNameKind, BTreeMap<String, String>>,
     merged_config: &OpaConfig,
+    user_info_fetcher_image: &str,
```
nightkr, line 545: This is currently always the operator image, not sure if we want to keep it that way or split into a sidecar image (would be kinda painful with the dockerfile workflow...)

rust/operator-binary/src/main.rs
```diff
@@ -36,13 +36,21 @@ pub mod built_info {
...
 struct Opts {
     #[clap(subcommand)]
-    cmd: Command<OpaRun>,
+    cmd: Cmd,
+}
+
+#[derive(Parser)]
+enum Cmd {
+    #[clap(flatten)]
+    Common(Command<OpaRun>),
```
nightkr, line 45: Old remnant, revert

rust/operator-binary/src/main.rs
```diff
@@ -36,13 +36,21 @@ pub mod built_info {
...
 }
 
 #[derive(clap::Parser)]
 struct OpaRun {
     #[clap(long, env)]
     opa_bundle_builder_clusterrole: String,
+    #[clap(long, env)]
+    operator_image: String,
```
nightkr, line 53: If we're moving the envvar into templating.. should this be moved into `ProductOperatorRun`? not sure

rust/user-info-fetcher/src/http_error.rs
```diff
@@ -0,0 +1,55 @@
+use std::sync::Arc;
```
nightkr, line 1: Could be moved out into some stackable-axum-commons library

tests/templates/kuttl/keycloak-user-info/10-install-opa.yaml.j2
```diff
@@ -0,0 +1,47 @@
...
+metadata:
+  name: test
+  labels:
+    opa.stackable.tech/bundle: "true"
+data:
+  test.rego: |
+    package test
+
+    userInfo(username) := http.send({"method": "POST", "url": "http://127.0.0.1:9476/user", "body": {"username": username}, "headers": {"Content-Type": "application/json"}, "raise_error": true}).body
+    currentUserInfo := userInfo(input.username)
```
nightkr, line 13: This should ideally be moved to some kind of "stackable library of rules", but we don't currently have a good way to deploy and manage that.

rust/operator-binary/src/main.rs
nightkr: 3aceb68612eb4f6285ba59f4ed3b447a773bb451

rust/operator-binary/src/controller.rs
nightkr: 0a684d9a7afe4b1855ccafc211696dc8e0967ed0

rust/operator-binary/src/main.rs
nightkr: 3aceb68612eb4f6285ba59f4ed3b447a773bb451

nightkr:
A few notes to self, but should largely be ready for review.

Tiltfile
soenkeliebau: Are you referring to the port forwards?

If so, maybe we can just assign fixed ports, like we already do for .... i dunno, something in templating I believe..Tilt itself maybe?

docs/modules/opa/examples/getting_started/opa.yaml
soenkeliebau: Document the userinfofetcher, or "none" backend?

rust/operator-binary/src/controller.rs
soenkeliebau: if its painful lets leave it as is for now

tests/templates/kuttl/keycloak-user-info/10-install-opa.yaml.j2
soenkeliebau: Is this the subsequent work with bundle-builder we talked about?

docs/modules/opa/examples/getting_started/opa.yaml
nightkr: The UIF. Obviously it should be documented *eventually*, it's more about at what stage we want to encourage people to start using it tbh.

tests/templates/kuttl/keycloak-user-info/10-install-opa.yaml.j2
nightkr: Yes.

deploy/helm/opa-operator/templates/deployment.yaml
NickLarsenNZ: Is the `image` attribute special, or should we be looking at prefixing it, eg: `stackable.tech/image`?

nightkr: It's arbitrary, that's probably a good call.

nightkr: 381ac91a56f91d8806f05ecf405c2d1862a00ac2

tests/templates/kuttl/keycloak-user-info/03-setup-keycloak.yaml.j2
```diff
@@ -0,0 +1,139 @@
...
+                    "implicitFlowEnabled": false,
+                    "directAccessGrantsEnabled": false,
+                    "serviceAccountsEnabled": true,
+                    "attributes": {
+                      "oidc.ciba.grant.enabled": "true",
+                      "oauth2.device.authorization.grant.enabled": false
+                    }
+                  }
+                  EOF
+                  # todo: is this needed if we are using Groups?
```
NickLarsenNZ: todo: remove this comment

docs/modules/opa/pages/usage-guide/user-info-fetcher.adoc
```diff
@@ -36,19 +39,72 @@ kind: Secret
...
+
+.About unencrypted HTTP
+[NOTE]
+===============================
+*NOTE* The User info fetcher serves endpoints over clear-text HTTP. 
+
+It is intended to only be accessed from the OPA Server via _localhost_ and to not be exposed outside of the Pod.
+===============================
+
+[source,python]
```
NickLarsenNZ, line 71: Set language to `rego` for when highlight.js finally supports it

NickLarsenNZ: Fixed in 5000c59652316090ba6e693e010acd2313645078

docs/modules/opa/pages/usage-guide/user-info-fetcher.adoc
```diff
@@ -36,19 +39,72 @@ kind: Secret
...
+<4> Refer to the applicable realm in your Keycloak server.
+
 Currently the following backends are supported:
 
 * <<_keycloak>>
 
 == Keycloak
 
+// todo: maybe this section should be under a Tutorial?
 Fetch groups and extra credentials, but not roles.
```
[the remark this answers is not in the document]
nightkr: Personally: I'd defer documentation until we have a client side API in place. I'd rather not have a bunch of users hard-code the HTTP requests.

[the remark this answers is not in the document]
nightkr: In the vein of "documentation causes ossification, and given our current maturity level I'd rather have no docs than stale docs".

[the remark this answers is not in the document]
NickLarsenNZ: Do you mean to revert the docs completely, or just this section?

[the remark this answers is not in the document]
nightkr: I would revert them entirely, or at least remove them from the nav hierarchy and add a very clear "UNSTABLE FEATURE" tag to the top.

deploy/helm/opa-operator/templates/deployment.yaml
NickLarsenNZ: @nightkr, should it be added into the [deployment template](https://github.com/stackabletech/operator-templating/blob/2b73b32beb75a63600e50a962c50b473fda62021/template/deploy/helm/%5B%5Boperator%5D%5D/templates/deployment.yaml.j2#L7-L8) directly, or into [the helper](https://github.com/stackabletech/operator-templating/blob/2ef5a917cce8d1710e1ce0d099947663f74b4428/template/deploy/helm/%5B%5Boperator%5D%5D/templates/_helpers.tpl#L40-L50)?

NickLarsenNZ: Or the [selectorLabels](https://github.com/stackabletech/operator-templating/blob/2ef5a917cce8d1710e1ce0d099947663f74b4428/template/deploy/helm/%5B%5Boperator%5D%5D/templates/_helpers.tpl#L52-L58)

NickLarsenNZ: My assumption is the deployment directly, because there are references to the annotations as mentioned in the comments below...

Tiltfile
NickLarsenNZ: Commented out in https://github.com/stackabletech/opa-operator/commit/8db6fc61a3cd7f493c78c03bfa3a71749467fd48#diff-c2ee8653e1d6b85f0aadf87cd438a9250806c052877248442be4d434cbc52425

deploy/helm/opa-operator/templates/deployment.yaml
NickLarsenNZ: I will also mention that `.Values.image.tag` is not defined: 

https://github.com/stackabletech/opa-operator/blob/b1644e4e49fd84879442fc9333cc947147faf47f/deploy/helm/opa-operator/values.yaml#L3-L6

Tiltfile
NickLarsenNZ: Done in https://github.com/stackabletech/operator-templating/pull/294

deploy/helm/opa-operator/templates/deployment.yaml
NickLarsenNZ: Done in https://github.com/stackabletech/operator-templating/pull/294

deploy/helm/opa-operator/templates/deployment.yaml
NickLarsenNZ: Done in https://github.com/stackabletech/operator-templating/pull/294

tests/templates/kuttl/keycloak-user-info/10-install-opa.yaml.j2
NickLarsenNZ: I've made a new issue to discuss this: https://github.com/stackabletech/opa-operator/issues/494

docs/modules/opa/examples/getting_started/opa.yaml
NickLarsenNZ: clusterConfig was removed in 61e58b2db3e3beefb759026fc1940f327ee35a85

rust/operator-binary/src/controller.rs
sbernauer: I agreed as well :D We worked hard to have product images that container everything to make product image selection easier

rust/user-info-fetcher/src/http_error.rs
nightkr: > Github makes it looks like you are commenting about use std::sync::Arc which is confusing, and I've assumed your comment refers to the whole file

Yes, correct. Goddamnit, github.. :/

nightkr: > I couldn't see a repo or a module by that name.

And yeah, this would be a new thing.

rust/operator-binary/src/main.rs
nightkr: Yes, it would be added to https://github.com/stackabletech/operator-rs/blob/f9a13a41b5637cb87b270600790b7b8604371d0a/src/cli.rs#L190-L200. properties.yaml (product-config) would not be affected.

tests/templates/kuttl/keycloak-user-info/10-install-opa.yaml.j2
```diff
@@ -0,0 +1,53 @@
...
+          currentUserInfoByUsername := userInfoByUsername(input.username)
+          currentUserInfoById := userInfoById(input.id)
+      ---
+      apiVersion: opa.stackable.tech/v1alpha1
+      kind: OpaCluster
+      metadata:
+        name: test-opa
+      spec:
+        image:
+          productVersion: "{{ test_scenario['values']['opa-latest'] }}"
```
sbernauer, line 30: Marker: Once merged, create a tag for UIF and use that in the docker-image

sbernauer (dismissed):
As discussed in call

rust/user-info-fetcher/src/main.rs
```diff
@@ -0,0 +1,273 @@
...
+#[serde(rename_all = "camelCase")]
+struct GroupRef {
+    name: String,
+}
+
+#[derive(Serialize, Clone, Debug)]
+#[serde(rename_all = "camelCase")]
+struct RoleRef {
+    name: String,
+}
```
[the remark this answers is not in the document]
sbernauer: 04b6953

rust/operator-binary/src/controller.rs
```diff
@@ -703,7 +728,35 @@ fn build_server_rolegroup_daemonset(
             ..Probe::default()
         });
 
-    let mut pb = PodBuilder::new();
+    if let Some(user_info) = &opa.spec.cluster_config.user_info {
+        cb_user_info_fetcher
+            .image(user_info_fetcher_image)
```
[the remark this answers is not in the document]
nightkr: The UIF comes from the _operator_ image, not the product image.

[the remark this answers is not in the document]
nightkr: While you could argue that the product image would be more correct.. the product image build process is enough of a mess as it is, and the strong dependency here is ultimately in making sure that the UIF and operator are compatible.

rust/user-info-fetcher/src/main.rs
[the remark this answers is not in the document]
nightkr: Not saying this is wrong here, but to explain myself.. the goal here was to ensure that we have room to add more fields here later (group/role IDs?).

rust/operator-binary/src/controller.rs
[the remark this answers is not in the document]
nightkr: This also isn't the only case of this, https://github.com/stackabletech/opa-operator/tree/spike/bundle-builder-v2 gives the bundle builder the same treatment.

tests/templates/kuttl/keycloak-user-info/10-install-opa.yaml.j2
nightkr: https://github.com/stackabletech/opa-operator/pull/433#discussion_r1422265176

rust/operator-binary/src/controller.rs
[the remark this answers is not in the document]
nightkr: There's also the DX side, testing repo-local changes are a lot less painful (especially when covered by tilt) than cross-repo changes.

rust/user-info-fetcher/src/http_error.rs
NickLarsenNZ: Given the new issue I created, can this comment be resolved? Or is it important enough to do this move now?

nightkr: I'd say that's fine.

rust/operator-binary/src/controller.rs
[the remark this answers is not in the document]
NickLarsenNZ: I agree that DX becomes a little more difficult now, but I think that could be good reason to move UIF out into its own repo with its own lifecycle (much like the bundle builder).

It is deployed by the operator, but it really isn't a component of the operator so it feels like it is only hanging around here because of DX.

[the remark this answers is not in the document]
nightkr: > I agree that DX becomes a little more difficult now, but I think that could be good reason to move UIF out into its own repo with its own lifecycle (much like the bundle builder).

How would DX be improved by that? Taking the bundle builder as a reference, you'd need to:

1. Make your change
2. Commit and push your change
3. Note the commit hash
4. Open the product Dockerfile (hope you have the right version of docker-images checked out already!)
5. Paste the commit hash into the Dockerfile
6. `bake`
7. Go make a coffee while you wait, because Docker's build caching is a sad joke
8. Go back to step 6, because you forgot to give the image a unique tag
9. Manually redeploy the `OpaCluster` with the new image version

And that's not even considering making sure that nightly operator versions match up with the right UIF versions, or that you now _also_ have to consider UIF config versioning, since you can't guarantee that the UIF uses the same version of the scheme as the operator anymore.

Tiltfile
NickLarsenNZ: Reverted in https://github.com/stackabletech/operator-templating/pull/297

But we are probably bringing it back as per slack huddle.

rust/user-info-fetcher/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+    http: reqwest::Client,
+    credentials: Arc<Credentials>,
+    user_info_cache: Cache<UserInfoRequest, UserInfo>,
+}
+
+#[derive(Debug)]
+struct Credentials {
+    // TODO: Find a better way of sharing behavior between different backends
+    username: String,
+    password: String,
```
nightkr, line 46:
```suggestion
    client_id: String,
    secret: String,
```

We might as well be truthful with the naming, and it would have to get refactored later anyway.

rust/user-info-fetcher/src/backend/keycloak.rs
```diff
@@ -0,0 +1,180 @@
...
+                    users_base_url
+                        .join(&req.id)
+                        .context(ConstructOidcEndpointPathSnafu)?,
+                )
+                .bearer_auth(&authn.access_token),
+            )
+            .await
+            .context(UserNotFoundByIdSnafu { user_id })?
+        }
+        UserInfoRequest::UserInfoRequestByName(req) => {
```
nightkr, line 146: Did we have any application that actually must query by username?

nightkr: If we do keep this around, should be `UserInfoRequestByUsername`.

rust/user-info-fetcher/src/backend/keycloak.rs
```diff
@@ -0,0 +1,180 @@
...
+            let user_name = &req.username;
```
nightkr, line 147:
```suggestion
            let username = &req.username;
```

Username != User's (display? legal?) name.

rust/user-info-fetcher/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+    });
+
+    let mut client_builder = ClientBuilder::new();
+
+    // TODO: I'm not so sure we should be doing all this keycloak specific stuff here.
+    // We could factor it out in the provider specific implementation (e.g. when we add LDAP support).
+    // I know it is for setting up the client, but an idea: make a trait for implementing backends
+    // The trait can do all this for a genric client using an implementation on the trait (eg: get_http_client() which will call self.uses_tls())
+    if let crd::Backend::Keycloak(keycloak) = &config.backend {
+        if keycloak.tls.uses_tls() && !keycloak.tls.uses_tls_verification() {
```
nightkr, line 129: This is more of a complaint about the op-rs API, but this feels really misuse-prone to me. If we want clients to branch from this then it would be preferable to have something like:

```
match tls.verification() {
    None => { ... }
    Some(TlsVerification::Custom { ca_mount_path }) => { ... }
    Some(TlsVerification::System) => { ... }
}
```

nightkr: To be clear, fixing that is way out of scope for this PR

rust/user-info-fetcher/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+            client_builder = client_builder.danger_accept_invalid_certs(true);
+        }
+        if let Some(tls_ca_cert_mount_path) = keycloak.tls.tls_ca_cert_mount_path() {
+            let mut buf = Vec::new();
+            File::open(tls_ca_cert_mount_path)
+                .await
+                .context(ReadCaCertSnafu)?
+                .read_to_end(&mut buf)
+                .await
+                .unwrap();
```
nightkr, line 139: Unwrap should be snafuified

rust/user-info-fetcher/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+
+#[derive(Debug, Clone)]
+struct AppState {
+    config: Arc<crd::Config>,
+    http: reqwest::Client,
+    credentials: Arc<Credentials>,
+    user_info_cache: Cache<UserInfoRequest, UserInfo>,
+}
+
+#[derive(Debug)]
```
nightkr, line 42: Ideally this shouldn't be `Debug`, or it should at least stub out `secret`.

rust/user-info-fetcher/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+enum GetUserInfoError {
+    #[snafu(display("failed to get user information from Keycloak"))]
+    Keycloak { source: backend::keycloak::Error },
+}
+
+impl http_error::Error for GetUserInfoError {
+    fn status_code(&self) -> hyper::StatusCode {
+        // todo: the warn here loses context about the scope in which the error occurred, eg: stackable_opa_user_info_fetcher::backend::keycloak
+        // Also, we should make the log level (warn vs error) more dynamic in the backend's impl `http_error::Error for Error`
+        tracing::warn!(error = ?self, "Error while processing request");
```
nightkr, line 212:
```suggestion
        tracing::warn!(error = &self as &dyn std::error::Error, "Error while processing request");
```

rust/user-info-fetcher/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+                                Some(id.clone())
+                            }
+                            _ => None,
+                        };
+                        let user_name = match &req {
+                            UserInfoRequest::UserInfoRequestByName(UserInfoRequestByName {
+                                username,
+                            }) => Some(username.clone()),
+                            _ => None,
+                        };
```
nightkr, line 245: Nit: useless clone

```suggestion
                        let (user_id, username) = match req {
                            UserInfoRequest::UserInfoRequestById(UserInfoRequestById { id }) => {
                                (Some(id), None)
                            }
                            UserInfoRequest::UserInfoRequestByName(UserInfoRequestByName {
                                username,
                            }) => (None, Some(username))
                        };
```

tests/templates/kuttl/keycloak-user-info/03-setup-keycloak.yaml.j2
```diff
@@ -0,0 +1,140 @@
...
+                    "directAccessGrantsEnabled": false,
+                    "serviceAccountsEnabled": true,
+                    "attributes": {
+                      "oidc.ciba.grant.enabled": "true",
+                      "oauth2.device.authorization.grant.enabled": false
+                    }
+                  }
+                  EOF
+
+                  kcadm.sh add-roles --uusername service-account-user-info-fetcher --rolename admin \$COMMON_FLAGS
```
nightkr, line 108: `--username`? Or is this a weird keycloakism?

tests/templates/kuttl/keycloak-user-info/02-install-keycloak.yaml.j2
```diff
@@ -0,0 +1,117 @@
...
+      metadata:
+        name: keycloak-tls-$NAMESPACE
+      spec:
+        backend:
+          autoTls:
+            ca:
+              autoGenerate: true
+              secret:
+                name: keycloak-tls-ca-$NAMESPACE
+                namespace: $NAMESPACE
```
nightkr, line 19: Do we really need to stub out the default vanilla `tls` secretclass?

nightkr (commented):
Only reviewed changes since 4d3613f04dd0002e4167cf52e95b4661444b1105 since.. something something impartiality

rust/user-info-fetcher/src/main.rs
NickLarsenNZ: I believe it was needed for Deserializing.

rust/user-info-fetcher/src/backend/keycloak.rs
NickLarsenNZ: > Did we have any application that actually must query by username?

Yes, Trino will when Superset is being used, because Superset is only sending through the `preferred_username` (when using OIDC).

rust/user-info-fetcher/src/main.rs
NickLarsenNZ: Agree. But even more ideally:

```rs
    client_id: String,
    client_secret: String,
```

nightkr: Sure.

rust/user-info-fetcher/src/backend/keycloak.rs
nightkr: Are we sure that Superset is sending the `preferred_username` (IdP username) and not the Superset username?

tests/templates/kuttl/keycloak-user-info/03-setup-keycloak.yaml.j2
NickLarsenNZ: It's an 'ism :/

rust/user-info-fetcher/src/backend/keycloak.rs
NickLarsenNZ: Fixed in 88c0e35

rust/user-info-fetcher/src/backend/keycloak.rs
NickLarsenNZ: You're right, my description wasn't accurate. 
Superset it sending through the username as it appears in the Superset user DB. When using OIDC with Keycloak, this comes from the `preferred_username` claim.

Trino is then sending whatever that is (it comes from the `x-trino-username` header).

rust/user-info-fetcher/src/main.rs
NickLarsenNZ: Fixed in e44102d

rust/user-info-fetcher/src/backend/keycloak.rs
nightkr: Right, so are we sure that these always match up? What happens when migrating a Superset from internal authn to OIDC? What happens if the username specified by `preferred_username` is already taken?

tests/templates/kuttl/keycloak-user-info/test-regorule.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python
+import requests
+import argparse
+import json
+
+# todo: make the test more comprehensive to check customAttributes
```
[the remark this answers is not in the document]
NickLarsenNZ: Will do this in a later PR

tests/templates/kuttl/keycloak-user-info/10-install-opa.yaml.j2
NickLarsenNZ: Not moving the UIF to the OPA image anymore

rust/user-info-fetcher/src/main.rs
NickLarsenNZ: Agree. Will close this here, and it will come up when it needs to.

rust/user-info-fetcher/src/main.rs
NickLarsenNZ: Fixed in cd7920d

rust/operator-binary/src/main.rs
NickLarsenNZ: I have created a branch and added the option, but I'm unsure if it should have a default empty value, and what the doc-comment description should be.

https://github.com/stackabletech/operator-rs/compare/main...product-operator-run-operator-image

rust/user-info-fetcher/src/backend/keycloak.rs
```diff
@@ -0,0 +1,179 @@
...
+            let users_url = users_base_url
+                .join(&format!("?username={user_name}&exact=true"))
+                .context(ConstructOidcEndpointPathSnafu)?;
+
+            send_json_request::<Vec<UserMetadata>>(
+                http.get(users_url).bearer_auth(&authn.access_token),
+            )
+            .await
+            .context(SearchForUserSnafu)?
+            .first() // FIXME: we should probably fail if there are more than one record
```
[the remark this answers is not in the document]
NickLarsenNZ: Fixed in c26bb0c91da74e10215e479e6c2ceac494c5627b

rust/user-info-fetcher/src/main.rs
NickLarsenNZ: Fixed in 59f106069224b13128a5d512490df25465c89263

rust/user-info-fetcher/src/backend/keycloak.rs
NickLarsenNZ: ignore this message... github bug that keeps moving focus to this :disappointed:

rust/user-info-fetcher/src/main.rs
NickLarsenNZ: I assume the `&` is not needed, as it results in:

```
211 |     fn status_code(&self) -> hyper::StatusCode {
    |                    - let's call the lifetime of this reference `'1`
...
215 |             error = &self as &dyn std::error::Error,
    |                     ^^^^^ cast requires that `'1` must outlive `'static`
```

I have removed the `?` and added the cast in ab945d2819816c1781d324ab04225a90fa19bdd2

rust/user-info-fetcher/src/main.rs
NickLarsenNZ: Fixed in f1b533222c8f5da331c392c342a6a969a105de43

tests/templates/kuttl/keycloak-user-info/02-install-keycloak.yaml.j2
NickLarsenNZ: Agreed, this ensures TLS is correctly trusted, and not by accident.

rust/user-info-fetcher/src/main.rs
NickLarsenNZ: Actually reverting that because...

```
 252 | ...                   id: user_id,
    |                           ^^^^^^^ expected `Option<String>`, found `Option<&String>`
```

rust/user-info-fetcher/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+                        let username = match &req {
+                            UserInfoRequest::UserInfoRequestByName(UserInfoRequestByName {
+                                username,
+                            }) => Some(username),
+                            _ => None,
+                        };
+                        Ok(UserInfo {
+                            id: user_id,
```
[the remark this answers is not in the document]
NickLarsenNZ: Fixed in 3e8a00d4b595e8a3515a36903031e43fc2acc281

rust/user-info-fetcher/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+                            username,
```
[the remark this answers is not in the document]
NickLarsenNZ: Fixed in 3e8a00d4b595e8a3515a36903031e43fc2acc281

rust/user-info-fetcher/src/main.rs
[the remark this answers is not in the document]
NickLarsenNZ: Fixed in 3e8a00d4b595e8a3515a36903031e43fc2acc281

rust/user-info-fetcher/src/main.rs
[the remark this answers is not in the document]
NickLarsenNZ: Fixed in 3e8a00d4b595e8a3515a36903031e43fc2acc281

rust/user-info-fetcher/src/backend/keycloak.rs
NickLarsenNZ: > Are we sure that these always match up?

When Impersonation is enabled on the Database Connection in Superset, Superset will send the logged in username in the `z-trino-username` header. When using OIDC, this originates from a claim, and in the case of Keycloak, is [hardcoded as `preferred_username`](https://github.com/dpgaspar/Flask-AppBuilder/blob/ba63c5c1629101dadf43e68a373402ee116f6ba9/flask_appbuilder/security/manager.py#L659-L671).

> What happens when migrating a Superset from internal authn to OIDC?
> What happens if the username specified by preferred_username is already taken?

No idea, my assumption would be as long as the applicable JWT claim (in the case of Keycloak, the `preferred_username` claim) matches the existing user, then it _is_ the same user. Just different authn mechanism. 

Would be interesting to test, but I believe that to be out-of-scope for this PR.

rust/operator-binary/src/main.rs
sbernauer: My suggestion:
```
    /// The full name of the operator image. This env var is set by the helm-charts and should
    /// therefore always be present.
    #[arg(long, env)]
    pub operator_image: String,
}
```

NickLarsenNZ: Cool, updated on the branch. Here's the PR: https://github.com/stackabletech/operator-rs/pull/702

rust/user-info-fetcher/src/main.rs
nightkr: You're right, since `self` is already `&Self` here.

tests/templates/kuttl/keycloak-user-info/02-install-keycloak.yaml.j2
nightkr: Sure.

rust/user-info-fetcher/src/backend/keycloak.rs
```diff
@@ -0,0 +1,190 @@
...
+
+#[derive(Snafu, Debug)]
+pub enum Error {
+    #[snafu(display("unable to log in (expired credentials?)"))]
+    LogIn { source: reqwest::Error },
+
+    #[snafu(display("unable to search for user"))]
+    SearchForUser { source: reqwest::Error },
+
+    #[snafu(display("user with id {user_id:?} was not found"))]
```
Techassi, line 19: Nit: We might want to come up with a rule on how we display "dynamic" data in an error message. I always feel like the "dynamic" data should be enclosed in some kind of quotes, for example: `user with id '123' was not found`.

Also, we should maybe aim for a consistent style of error messages like:

- "Unable to find ..." or
- "Failed to find ..."

Techassi: Mixing "unable", "failed", and a completely different form of sentence in one type of error seems a little weird.

rust/user-info-fetcher/src/http_error.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use axum::{response::IntoResponse, Json};
+use hyper::StatusCode;
+use serde::Serialize;
+
+pub trait Error: std::error::Error {
```
Techassi, line 7: We might want to call that `ErrorExt`, as it extends the `std::error::Error` trait.

rust/user-info-fetcher/src/http_error.rs
```diff
@@ -0,0 +1,55 @@
...
+    fn status_code(&self) -> StatusCode;
+}
+impl<T: Error> Error for Arc<T> {
+    fn status_code(&self) -> StatusCode {
+        let inner: &T = self;
+        inner.status_code()
+    }
+}
+
+pub struct JsonResponse<E> {
```
Techassi, line 17: I feel like `JsonResponse` might not be the best name for how the struct is used (indicating an error occurred). How about `JsonErrorResponse` or something similar?

Techassi (changes requested):
Partial review, I will look into the rest (the bulk of `main.rs`) asap.

rust/user-info-fetcher/src/http_error.rs
NickLarsenNZ: And remove the generic? Yeah could do. This should all move to operator-rs anyway.

rust/crd/src/user_info_fetcher.rs
```diff
@@ -0,0 +1,86 @@
...
+    #[derivative(Default(value = "Cache::default_entry_time_to_live()"))]
+    #[serde(default = "Cache::default_entry_time_to_live")]
+    pub entry_time_to_live: Duration,
+}
+
+impl Cache {
+    const fn default_entry_time_to_live() -> Duration {
+        Duration::from_minutes_unchecked(1)
+    }
+}
```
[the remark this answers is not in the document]
nightkr: The caching is implemented in the UIF crate.

rust/user-info-fetcher/src/http_error.rs
NickLarsenNZ: Oh actually the generic E looks like it needs to stay based on the other lines below

rust/user-info-fetcher/src/backend/keycloak.rs
NickLarsenNZ: I agree with quoting variable data.
And yeah, I think "Failed" should be reserved for actual failures (eg: error 500 looking up user, rather than 404, user not found). Maybe we do need some guidelines written.

nightkr: `{:?}` will quote string values. But agreed that it should be written down somewhere.

nightkr: Also agreed with @nicklarsennz about failed.

rust/user-info-fetcher/src/http_error.rs
nightkr: Conventionally `FooExt` is for a trait adding new operations to `Ext`. A more appropriate name would be `HttpEndpointError`, but that's a mouthful.

rust/user-info-fetcher/src/http_error.rs
nightkr: The intention here is that you'd use it as `error::JsonResponse`. And yeah the generic is inherent to both the data storage and the serde impls.

rust/user-info-fetcher/src/backend/keycloak.rs
nightkr: > When using OIDC, this originates from a claim, and in the case of Keycloak, is [hardcoded as preferred_username](https://github.com/dpgaspar/Flask-AppBuilder/blob/ba63c5c1629101dadf43e68a373402ee116f6ba9/flask_appbuilder/security/manager.py#L659-L671).

I'll need to look deeper, but often that sort of thing is just the _suggested_ data to prefill the sign-up form. And as you can see, in that file it's just as often [hard-coded to *not* match the backend username](https://github.com/dpgaspar/Flask-AppBuilder/blob/ba63c5c1629101dadf43e68a373402ee116f6ba9/flask_appbuilder/security/manager.py#L645).

> Would be interesting to test, but I believe that to be out-of-scope for this PR.

Yes and no. I'd rather not add an API path that encourages people to do the wrong thing... :/

rust/operator-binary/src/controller.rs
```diff
@@ -751,6 +801,23 @@ fn build_server_rolegroup_daemonset(
             .build(),
     );
 
+    if let Some(user_info) = &opa.spec.cluster_config.user_info {
```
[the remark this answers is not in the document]
sbernauer: At least first part: fcbc5ea

rust/crd/src/user_info_fetcher.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// Port of the identity provider. If TLS is used defaults to `443`, otherwise to `80`.
+    pub port: Option<u16>,
+
+    /// Root HTTP path of the identity provider. Defaults to `/`.
+    #[serde(default = "default_root_path")]
+    pub root_path: String,
+
+    /// Use a TLS connection. If not specified no TLS will be used.
+    #[serde(flatten)]
+    pub tls: TlsClientDetails,
```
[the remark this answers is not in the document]
sbernauer: I assume you are refering to https://github.com/stackabletech/operator-rs/blob/38cb237796d10f8f3ba8d9db1e78f444e0306dd1/src/commons/authentication/oidc.rs#L40?
The problem here is that we not only have 3 additional fields, but also op-rs has 3 fields that we don't need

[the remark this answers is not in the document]
sbernauer: But at least we use that struct to call helper functions on :sweat: 
https://github.com/stackabletech/opa-operator/blob/c26bb0c91da74e10215e479e6c2ceac494c5627b/rust/user-info-fetcher/src/backend/keycloak.rs#L103-L116

rust/operator-binary/src/controller.rs
```diff
@@ -53,17 +54,20 @@ use stackable_operator::{
         operations::ClusterOperationsConditionBuilder,
     },
     time::Duration,
-    utils::COMMON_BASH_TRAP_FUNCTIONS,
+};
+use stackable_operator::{
+    k8s_openapi::api::core::v1::SecretVolumeSource,
+    product_config_utils::validate_all_roles_and_groups_config, utils::COMMON_BASH_TRAP_FUNCTIONS,
```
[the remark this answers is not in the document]
sbernauer: I don't think this was intentional. Merged in 515d3f6

rust/operator-binary/src/controller.rs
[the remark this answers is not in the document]
Techassi: Further improvement: 6610346

rust/crd/src/user_info_fetcher.rs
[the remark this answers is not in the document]
Techassi: OKay, then nevermind... Just a thought.

rust/crd/src/user_info_fetcher.rs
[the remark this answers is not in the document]
Techassi: Ah yeah, this is only used for configuration. Marking as resolved.

rust/user-info-fetcher/src/backend/keycloak.rs
NickLarsenNZ: improved error messages in 33e8893

rust/user-info-fetcher/src/http_error.rs
NickLarsenNZ: It does seem to be extending `std::error:Error` in the same way that the SubscriberExt trait extends Subscriber in the tracing crate.
https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/trait.SubscriberExt.html

So maybe it should be ErrorExt

rust/user-info-fetcher/src/http_error.rs
Techassi: I don't have any super strong opinions on that, so we can move forward with it for now.

rust/user-info-fetcher/src/http_error.rs
nightkr: `SubscriberExt` adds a *consumer-facing* convenience API on top of `Subscriber`'s *implementer-facing* API. You can see that [it has a single empty blanket impl](https://docs.rs/tracing-subscriber/0.3.18/src/tracing_subscriber/layer/mod.rs.html#1898).

The same pattern also holds elsewhere. [`futures::Future`](https://docs.rs/futures/0.3.29/futures/future/trait.Future.html) defines the base implementer-facing API (these days via `std`), [`FutureExt`](https://docs.rs/futures/0.3.29/futures/future/trait.FutureExt.html) defines the consumer-facing API as a blanket impl. [`TryFuture`](https://docs.rs/futures/0.3.29/futures/future/trait.TryFuture.html) defines a subset of `Future`s as fallible, [`TryFutureExt`](https://docs.rs/futures/0.3.29/futures/future/trait.TryFutureExt.html) provides the consumer-facing APIs that this implies. `TryFuture` itself is _not_ an `Ext` trait, because it carries its own semantics on top of the base trait.

`http_error::Error` is the opposite, it's a way for the error type to communicate _more_ information to the consumer, which isn't given by the std `Error`. There is no blanket impl, and there would be no way to provide a useful one.

rust/operator-binary/src/controller.rs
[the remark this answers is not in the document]
Techassi: We can resolve this for now, if we don't want to move the code into its own function.

rust/operator-binary/src/controller.rs
```diff
@@ -703,53 +741,97 @@
...
+                            secret_name: Some(keycloak.client_credentials_secret.clone()),
+                            ..Default::default()
+                        })
+                        .build(),
+                );
+                cb_user_info_fetcher.add_volume_mount(
+                    USER_INFO_FETCHER_CREDENTIALS_VOLUME_NAME,
+                    USER_INFO_FETCHER_CREDENTIALS_DIR,
+                );
+                keycloak
```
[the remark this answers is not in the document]
NickLarsenNZ: I'll look into this one

rust/operator-binary/src/controller.rs
```diff
@@ -703,53 +741,97 @@ fn build_server_rolegroup_daemonset(
...
```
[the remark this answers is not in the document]
NickLarsenNZ: Fixed in c20baa1388664ba581d267e98522aae6177353c8

rust/operator-binary/src/controller.rs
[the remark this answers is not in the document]
NickLarsenNZ: Fixed in c20baa1388664ba581d267e98522aae6177353c8

rust/user-info-fetcher/src/http_error.rs
NickLarsenNZ: Thanks for the explanation. So this will stay as is (and will be moved as part of #493)

rust/user-info-fetcher/src/backend/keycloak.rs
NickLarsenNZ: Yeah, I have no idea what is going on with that openshift user example. But that seems far beyond out control.

> I'd rather not add an API path that encourages people to do the wrong thing... :/

Sure, but I think we are at the mercy of the product implementations. Superset uses Flask-Appbuilder with the already mentioned oddities baked in. Trino only allows a single piece of data to be sent for impersonation. I'm not really sure there is much we can do on our side to discourage what we think is the wrong thing.

Do you have any suggestions?

rust/operator-binary/src/controller.rs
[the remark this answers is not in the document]
NickLarsenNZ: I don't have a comprehensive answer on how DX could be improved, but I'm sure it could in the same way other things have been improved. Naivly, I would say that when doing user-info-fetcher development, Tilt can go and supply the correct image for deployment by the opa-operator.

But in this case, we have agreed to keep it where it is, so I believe we just need to revert a change and this can be resolved.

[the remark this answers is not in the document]
NickLarsenNZ: Reverted in 6dcc44cad25df9c2a797727efa84a2e4bd637af7 and the additional `.image` call added at @sbernauer's request.

rust/operator-binary/src/main.rs
NickLarsenNZ: That PR was closed as it was deemed too premature.

rust/operator-binary/src/controller.rs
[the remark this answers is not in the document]
sbernauer: Thanks!

rust/user-info-fetcher/src/backend/keycloak.rs
NickLarsenNZ: Update: we had a discussion to give out opinions and try to reach a resolution.

The outcome of it was:
- Merge as is, and (optionally?) create a new issue for any renaming of fields (for example, `username` -> `effective_username` in the UIF response data).
- Leave docs hidden for now. Merging this now is not a promise of stability or release (even if the code is technically part of the next release because it is in the main branch).
- Have a discussion about branching philosophy as there appears to be multiple trains of thought on it, each with their own advantages and disadvantages.

_I hope I have captured this correctly, but I will leave it open for @nightkr to confirm, or add/correct anything I got wrong._

nightkr: That's my understanding as well.

NickLarsenNZ: Cool, I'll mark as resolved and merge it.

rust/user-info-fetcher/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+description = "Fetches user metadata from directory services"
+version.workspace = true
+authors.workspace = true
+license.workspace = true
+edition.workspace = true
+repository.workspace = true
+publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
```
sbernauer, line 12:
```suggestion
```

sbernauer (dismissed):
Gone through all changes again (including the new labels mechanism) and LGTM.
Many thanks for everyone that has participated, it was a great group effort!

sbernauer:
We need the fixes from https://github.com/stackabletech/operator-templating/pull/306, but with these the kuttl test passes again

sbernauer (approved):
Test https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/opa-operator-it-custom/85 passed :rocket:

NickLarsenNZ (approved):
Tests pass, good to go. Thanks everyone for your contributions

PR #510, opened 2024-01-31
feat(UserInfoFetcher): better errors for HTTP errors

### LINKED ISSUE #509: User Info fetcher: Improve http request error output
# Problem

Currently, we raise an error if the response to an HTTP request is 4xx or 5xx.  We only know which part of the code the User Info Fetcher failed on. We have no HTTP response body in the error to give a hint to the problem. Eg: Is it a WAF or proxy issue, or are we reaching the idP but some configuration is wrong?

https://github.com/stackabletech/opa-operator/blob/a4a6f4490b8e10e937606e78b66914cd873516c4/rust/user-info-fetcher/src/backend/keycloak.rs#L11-L14

The difficulty in adding the extra indfoamtion is that the Reqwest library consumes the response when calling methods like `.text()` (or `.json()`), or `error_for_status()`, and we need a combination of both methods.

# Suggested fix

Don't make use of `error_for_status()` (which simply checks for 4xx or 5xx errors) and instead manually check the status code, and then either response with the `json()`, or produce a Snafu error containing the `text()` (which is fallible, so consideration should be made how to handle that).

# Acceptance criteria

- [ ] Snafu errors for `util::send_json_request`
- [ ] 4xx and 5xx still produce an error, but also contain any body output
  - Consider what happens if the decoding fails, or is very long.

fhennig:
# Description

fixes #509

```[tasklist]
# Reviewer
- [ ] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/style-guide).
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/rust/user-info-fetcher/src/util.rs
+++ b/rust/user-info-fetcher/src/util.rs
@@ -15,7 +15,7 @@ pub enum Error {
     HttpErrorResponse {
         status: StatusCode,
         url: Url,
-        text: String
+        text: String,
     },
 
     #[snafu(display("response was an HTTP error with undecodable text"))]
@@ -41,21 +41,18 @@ pub async fn send_json_request<T: DeserializeOwned>(req: RequestBuilder) -> Resu
 /// reqwest error does not give any response body context.
 async fn error_for_status(response: Response) -> Result<Response, Error> {
     let status = response.status();
-    // good response
-    if status.is_success() || status.is_informational() || status.is_redirection() {
-        Ok(response)
-    }
-    // error response branch -> get the text and raise an error
-    else {
+    if status.is_client_error() || status.is_server_error() {
         let url = response.url().to_owned();
         match response.text().await {
-            Ok(text) => HttpErrorResponseSnafu { status, url, text }.fail(),
+            Ok(text) => HttpErrorResponseSnafu { status, url, text }.fail()?,
             Err(encoding_error) => HttpErrorResponseUndecodableTextSnafu {
                 status,
                 url,
                 encoding_error,
             }
-            .fail(),
+            .fail()?,
         }
+    } else {
+        Ok(response)
     }
 }
--- a/rust/user-info-fetcher/src/util.rs
+++ b/rust/user-info-fetcher/src/util.rs
@@ -44,16 +44,15 @@ async fn error_for_status(response: Response) -> Result<Response, Error> {
     let status = response.status();
     if status.is_client_error() || status.is_server_error() {
         let url = response.url().to_owned();
-        match response.text().await {
-            Ok(text) => HttpErrorResponseSnafu { status, url, text }.fail()?,
+        return match response.text().await {
+            Ok(text) => HttpErrorResponseSnafu { status, url, text }.fail(),
             Err(encoding_error) => HttpErrorResponseUndecodableTextSnafu {
                 status,
                 url,
                 encoding_error,
             }
-            .fail()?,
-        }
-    } else {
-        Ok(response)
+            .fail(),
+        };
     }
+    Ok(response)
 }
--- a/rust/user-info-fetcher/src/util.rs
+++ b/rust/user-info-fetcher/src/util.rs
@@ -36,9 +36,10 @@ pub async fn send_json_request<T: DeserializeOwned>(req: RequestBuilder) -> Resu
     Ok(result)
 }
 
-/// takes a Response and checks whether it is an error. If so, parse the reqwest Error
-/// and create our own error type with more context added. We do this because the plain
-/// reqwest error does not give any response body context.
+/// Wraps a Response into a Result. If there is an HTTP Client or Server error,
+/// extract the HTTP body (if possible) to be used as context in the returned Err.
+/// This is done this because the `Response::error_for_status()` method Err variant
+/// does not contain this information.
 async fn error_for_status(response: Response) -> Result<Response, Error> {
     let status = response.status();
     if status.is_client_error() || status.is_server_error() {
--- a/rust/user-info-fetcher/src/util.rs
+++ b/rust/user-info-fetcher/src/util.rs
@@ -1,8 +1,28 @@
 use reqwest::RequestBuilder;
 use serde::de::DeserializeOwned;
+use snafu::Snafu;
+
+// TODO write an error type in here
+
+#[derive(Snafu, Debug)]
+pub enum Error {
+    #[snafu(display("failed to execute request"))]
+    HttpRequest { source: reqwest::Error },
+
+    #[snafu(display("failed to parse json response"))]
+    ParseJson { source: reqwest::Error },
+
+    #[snafu(display("response was an http error"))]
+    HttpErrorResponse { source: reqwest::Error },
+
+    #[snafu(display("response was an http error: {text}"))]
+    Something { text: String },
+}
 
 pub async fn send_json_request<T: DeserializeOwned>(
     req: RequestBuilder,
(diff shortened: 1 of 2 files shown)


rust/user-info-fetcher/src/util.rs
```diff
@@ -1,8 +1,54 @@
...
+    else {
+        match response.text().await {
+            Ok(text) => HttpErrorResponseSnafu { status, text }.fail(),
+            Err(encoding_error) => HttpErrorResponseUndecodableTextSnafu {
+                status,
+                encoding_error,
+            }
+            .fail(),
+        }
+    }
```
NickLarsenNZ, line 53: **thought** (non-blocker): I would probably have flipped the logic to keep it closer to how reqwest does it in `error_for_status()`.

I would also avoid the `else` branch, as per:
- [Else considered smelly](https://wiki.c2.com/?ElseConsideredSmelly=)
- [Return early pattern](https://medium.com/swlh/return-early-pattern-3d18a41bba8)

```rust
    let status = response.status();

    // http client/server error response -> get the text and raise an error
    if status.is_client_error() || is_server_error {
        match response.text().await {
            Ok(text) => return HttpErrorResponseSnafu { status, text }.fail(),
            Err(encoding_error) => return HttpErrorResponseUndecodableTextSnafu {
                status,
                encoding_error,
            }
            .fail(),
        }
    }

    Ok(response)
```

rust/user-info-fetcher/src/util.rs
```diff
@@ -1,8 +1,54 @@
...
+    // check for client or server errors
+    let non_error_response = get_non_error_response(response).await?;
+    // parse the result
+    let result = non_error_response.json().await.context(ParseJsonSnafu)?;
+    Ok(result)
+}
+
+/// takes a Response and checks whether it is an error. If so, parse the reqwest Error
+/// and create our own error type with more context added. We do this because the plain
+/// reqwest error does not give any response body context.
```
NickLarsenNZ, line 36:
```suggestion
/// Wraps a Response into a Result. If there is an HTTP Client or Server error, extract the HTTP body (if possible) to be used as context in the returned Err. 
/// This is done this because the `Response::error_for_status()` method Err variant does not contain this information.
```

rust/user-info-fetcher/src/util.rs
NickLarsenNZ: _I think instead of the explicit returns, a `?` can be used after the `match`._

NickLarsenNZ (dismissed):
LGTM, left some comments that should be non-blockers.

Also trying to use [Conventional Comments](https://conventionalcomments.org/) to make the distinction between actionable and non-actionable comments are (while still sharing thoughts).

rust/user-info-fetcher/src/util.rs
Techassi: I fully agree on using the protective / guard pattern as much as possible. Returning as early as possible is always nice to have!

rust/user-info-fetcher/src/util.rs
NickLarsenNZ: _sorry, the suggestion feature and line breaks conflict_

rust/user-info-fetcher/src/util.rs
```diff
@@ -1,8 +1,54 @@
...
+    ParseJson { source: reqwest::Error },
+
+    #[snafu(display("response was an HTTP error: {text}"))]
+    HttpErrorResponse { status: StatusCode, text: String },
+
+    #[snafu(display("response was an HTTP error with undecodable text"))]
+    HttpErrorResponseUndecodableText {
+        status: StatusCode,
+        encoding_error: reqwest::Error,
+    },
```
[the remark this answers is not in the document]
fhennig: Alright, makes sense! I found the `url()` on the Response object and I'll add that in, but I don't think the HTTP method is there: https://docs.rs/reqwest/latest/reqwest/struct.Response.html

I did see `headers` though, maybe those are interesting?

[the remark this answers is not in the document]
fhennig: I added the `url` here: https://github.com/stackabletech/opa-operator/pull/510/commits/8dc7a262bd7407d712cd5f5e60605cf8e49f3587

rust/user-info-fetcher/src/util.rs
```diff
@@ -1,8 +1,54 @@
...
+    let non_error_response = get_non_error_response(response).await?;
+    // parse the result
+    let result = non_error_response.json().await.context(ParseJsonSnafu)?;
+    Ok(result)
+}
+
+/// takes a Response and checks whether it is an error. If so, parse the reqwest Error
+/// and create our own error type with more context added. We do this because the plain
+/// reqwest error does not give any response body context.
+async fn get_non_error_response(response: Response) -> Result<Response, Error> {
```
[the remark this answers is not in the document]
fhennig: Good point, I think `error_for_status` is actually a good idea: https://github.com/stackabletech/opa-operator/pull/510/commits/8dc7a262bd7407d712cd5f5e60605cf8e49f3587

rust/user-info-fetcher/src/util.rs
fhennig: I flipped the logic but couldn't get rid of the `else` when using `?` as well. I decided I'd rather have `?` and `else` than `return`s and no else: https://github.com/stackabletech/opa-operator/pull/510/commits/8d4427b495ed21fc2edd89cc05460a2207f2bb2b

rust/user-info-fetcher/src/util.rs
fhennig: done: https://github.com/stackabletech/opa-operator/pull/510/commits/e0a5b9284263cf8cea8a7ceb6e1b5a8aaa4be8e8

rust/user-info-fetcher/src/util.rs
NickLarsenNZ: It should work. The `?` shouldn't be after fail() I don't thing, but rather after the match. Because the `match` will always return a Result::Err (because of what fail is doing in both arms):

```rust
    let status = response.status();
    if status.is_client_error() || status.is_server_error() {
        let url = response.url().to_owned();
        match response.text().await {
            Ok(text) => HttpErrorResponseSnafu { status, url, text }.fail(),
            Err(encoding_error) => HttpErrorResponseUndecodableTextSnafu {
                status,
                url,
                encoding_error,
            }
            .fail(),
        }?
    }

    Ok(response)
```

Maybe I'm missing something

NickLarsenNZ: Oh, a use after move seems to occur, but I think this works (prepending the match with an early return):

```rust
async fn error_for_status(response: Response) -> Result<Response, Error> {
    let status = response.status();
    if status.is_client_error() || status.is_server_error() {
        let url = response.url().to_owned();
        return match response.text().await {
            Ok(text) => HttpErrorResponseSnafu { status, url, text }.fail(),
            Err(encoding_error) => HttpErrorResponseUndecodableTextSnafu {
                status,
                url,
                encoding_error,
            }
            .fail(),
        };
    }

    Ok(response)
}
```

fhennig: https://github.com/stackabletech/opa-operator/pull/510/commits/587db50beb6b5d15a943b6989c2490697706577e

Alright, fine with using a return instead of the question mark

NickLarsenNZ (approved):
Looks good. Thanks

PR #540, opened 2024-03-11
fix(tests): Add service account to Keycloak deployment

razvan:


tests/test-definition.yaml
```diff
@@ -31,6 +31,7 @@ tests:
     dimensions:
       - opa-latest
       - keycloak
+      - openshift
```
sbernauer, line 34: LGTM, but I think we should comment out line 15, WDYT?

sbernauer (approved):
Thanks!

tests/test-definition.yaml
razvan: Done. I didn't see it. Thanks.

PR #557, opened 2024-04-26
docs: Promote user-info-fetcher docs to navgation

sbernauer:
# Description

Given that we are using it extensively already (e.g. in the e2e-security demo) and me handing out the link to multiple customers, I would like to promote https://docs.stackable.tech/home/nightly/opa/usage-guide/user-info-fetcher.html to include it in the navigation

```[tasklist]
# Author
- [ ] CRD changes approved
- [ ] CRD documentation for all fields, following the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs-style-guide).
- [ ] Changes need to be "offline" compatible
```

```[tasklist]
# Reviewer
- [ ] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs-style-guide).
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
- [ ] [Roadmap](https://github.com/orgs/stackabletech/projects/25/views/1) has been updated
```


Maleware (dismissed):
LGTM! Thank you :)

sbernauer:
Giving @nightkr and @NickLarsenNZ a chance to comment...

nightkr:
Have we addressed the client API side yet? We shouldn't have been shipping it to demos _or_ customers before that, and the linked docs definitely don't.

sbernauer:
Are you talking about the API described in https://docs.stackable.tech/home/nightly/opa/usage-guide/user-info-fetcher.html#_user_info_fetcher_api?
Currently you can send an username or id and get all infos we can collect back. What exactly do you want to address, would you like to change the API?

nightkr:
The UIF provides the HTTP API, but we should wrap that in an OPA rule we provide, rather than have everyone hard-code the HTTP stuff themselves.

fhennig:
I don't think we have that yet. I agree that it would be good to have. Could we simply have a `uif` rego package that we deploy when the UIF is enabled? and users can import the package in their own code?

I think we can still do this 'promotion' here and do a follow-up PR to adress this. it's all in nightly anyways, but we need to prioritise the follow-up then

sbernauer:
Yeah, I can see the point of providing a regorule in addition to the HTTP API. The question is if we should block the docs on the API until the rego rules are there :thinking:

nightkr:
> I don't think we have that yet. I agree that it would be good to have. Could we simply have a uif rego package that we deploy when the UIF is enabled? and users can import the package in their own code?

Yes, basically this.

> Yeah, I can see the point of providing a regorule in addition to the HTTP API. The question is if we should block the docs on the API until the rego rules are there 🤔

We should, because the current HTTP API *is not public or stable*. The uif package is the stable API surface, the HTTP API is just an implementation detail for how to get there.

fhennig (dismissed):
blocked by #558

sbernauer:
I think the change is trivial and I would close this PR until everyone thinks we are ready to include the docs in the navigation

sbernauer:
Reopening according to the decision in the daily, as we where not able to ship https://github.com/stackabletech/opa-operator/issues/558 in time for the 23.07 release

NickLarsenNZ (dismissed):
LGTM

docs/modules/opa/pages/usage-guide/user-info-fetcher.adoc
```diff
@@ -62,6 +62,16 @@ NOTE: The OAuth2 Client in Keycloak must be given the `view-users` _Service Acco
 // TODO: Document how to use it in OPA regorules, e.g. to authorize based on group membership
 == Example rego rule
 
+[NOTE]
+.User-facing API & API stability
+====
+As of the SDP release 24.07 we document an example rego rule here making an HTTP call.
+However, our plan is that the user-facing API of the SPD is *not* the HTTP API of user-info-fetcher, but instead regorules that will automatically be shipped to the OPA server.
+This enables us e.g. to make breaking changes to the HTTP API while still keeping the rego rules API stable.
```
Techassi, line 70:
```suggestion
This enables us to make underlying (for example breaking) changes to the HTTP API while keeping the rego rules API stable.
```

Techassi (changes requested):
Some suggestions.

Techassi (approved):

PR #698, opened 2025-03-19
Use service scope for user-info-fetcher AD kerberos keytab

### LINKED ISSUE #680: UserInfoFetcher Active Directory backend should use service scope for Kerberos
Currently it uses pod+node, which is both unclear/insecure (since the node principal will be shared with anything else running on that node) and problematic for k8sSearch secrets (since the pod name is unpredictable).

Instead, we should use a common service scope that is shared between all pods.

This is technically breaking since we change the scopes k8sSearch users would need to specify, but in practice it's currently unusable with k8sSearch anyway.

nightkr:
# Description

Fixes #680

```[tasklist]
# Author
- [X] Helm chart can be installed and deployed operator works
- [X] Integration tests passed (for non trivial changes)
```

```[tasklist]
# Reviewer
- [ ] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
- [ ] [Roadmap](https://github.com/orgs/stackabletech/projects/25/views/1) has been updated
```

### DIFF
--- a/tests/templates/kuttl/ad-user-info/10-install-opa.yaml.j2
+++ b/tests/templates/kuttl/ad-user-info/10-install-opa.yaml.j2
@@ -56,14 +56,5 @@ commands:
           config:
             logging:
               enableVectorAgent: {{ lookup('env', 'VECTOR_AGGREGATOR') | length > 0 }}
-          podOverrides:
-            spec:
-              volumes:
-              - name: kerberos
-                ephemeral:
-                  volumeClaimTemplate:
-                    metadata:
-                      annotations:
-                        secrets.stackable.tech/scope: service=opa
           roleGroups:
             default: {}
--- a/tests/templates/kuttl/ad-user-info/10-install-opa.yaml.j2
+++ b/tests/templates/kuttl/ad-user-info/10-install-opa.yaml.j2
@@ -1,0 +1,69 @@
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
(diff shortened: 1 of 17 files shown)


tests/templates/kuttl/ad-user-info/10-install-opa.yaml.j2
```diff
@@ -0,0 +1,69 @@
...
+              enableVectorAgent: {{ lookup('env', 'VECTOR_AGGREGATOR') | length > 0 }}
+          podOverrides:
+            spec:
+              volumes:
+              - name: kerberos
+                ephemeral:
+                  volumeClaimTemplate:
+                    metadata:
+                      annotations:
+                        secrets.stackable.tech/scope: service=opa
```
sbernauer, line 67: I'm a bit confused why you are using podOverrides here. Shouldn't the operator set it to this exact value?

nightkr: The answer is that I'm a dumbass who used an old test I wrote ages ago. Will fix (and retest).

nightkr: c493329960e69d2895f6fcce17ef89cf8a44590c

sbernauer: Makes perfect sense, thanks!

sbernauer (approved):
Nice fix, thanks!

sbernauer (approved):

PR #712, opened 2025-04-14
feat: Add experimental Entra user-info-fetcher backend

maltesander:
# Description

- adds expermimental Entra UIF backend

CRD

```rust
pub struct EntraBackend {
    /// Hostname of the token provider, defaults to `login.microsoft.com`.
    #[serde(default = "entra_default_token_hostname")]
    pub token_hostname: HostName,

    /// Hostname of the user info provider, defaults to `graph.microsoft.com`.
    #[serde(default = "entra_default_user_info_hostname")]
    pub user_info_hostname: HostName,

    /// Port of the identity provider. If TLS is used defaults to `443`, otherwise to `80`.
    pub port: Option<u16>,

    /// The Microsoft Entra tenant ID.
    pub tenant_id: String,

    /// Use a TLS connection. Should usually be set to WebPki.
    // We do not use the flattened `TlsClientDetails` here since we cannot
    // default to WebPki using a default and flatten
    // https://github.com/serde-rs/serde/issues/1626
    // This means we have to wrap `Tls` in `TlsClientDetails` to use its
    // method like `uses_tls()`.
    #[serde(default = "default_tls_web_pki")]
    pub tls: Option<Tls>,

    /// Name of a Secret that contains client credentials of an Entra account with
    /// permissions `User.ReadAll` and `GroupMemberShip.ReadAll`.
    ///
    /// Must contain the fields `clientId` and `clientSecret`.
    pub client_credentials_secret: String,
}
```

And the usage:

```yaml
---
apiVersion: v1
kind: Secret
metadata:
  name: user-info-fetcher-client-credentials
stringData:
  clientId: foo
  clientSecret: bar
---
apiVersion: opa.stackable.tech/v1alpha1
kind: OpaCluster
metadata:
  name: opa
spec:
  image:
    productVersion: "1.0.1"
  clusterConfig:
    userInfo:
      backend:
        experimentalEntra:
          tenantId: foo-bar
          clientCredentialsSecret: user-info-fetcher-client-credentials
  servers:
    roleGroups:
      default: {}
```


deploy/helm/opa-operator/crds/crds.yaml
```diff
@@ -137,6 +139,82 @@ spec:
                                 - kerberosSecretClassName
                                 - ldapServer
                               type: object
+                            experimentalEntraBackend:
```
sbernauer, line 142: I prefer `experimentalEntra` over `experimentalEntraBackend`, as we don't suffix `Backend` anywhere else

rust/user-info-fetcher/src/backend/entra.rs
```diff
@@ -0,0 +1,267 @@
...
+    access_token: String,
+}
+
+/// The minimal structure of [UserRepresentation] that is returned by [`/users`][users] and [`/users/{id}`][user-by-id].
+/// <div class="warning">Some fields, such as `groups` are never present. See [keycloak/keycloak#20292][issue-20292]</div>
+///
+/// [users]: https://www.keycloak.org/docs-api/22.0.1/rest-api/index.html#_get_adminrealmsrealmusers
+/// [user-by-id]: https://www.keycloak.org/docs-api/22.0.1/rest-api/index.html#_get_adminrealmsrealmusersid
+/// [UserRepresentation]: https://www.keycloak.org/docs-api/22.0.1/rest-api/index.html#UserRepresentation
+/// [issue-20292]: https://github.com/keycloak/keycloak/issues/20294
```
sbernauer, line 57: Maybe I'm stupid, but why are there lot's of Keycloak links in an Entra implementation? :see_no_evil:

rust/user-info-fetcher/src/backend/entra.rs
```diff
@@ -0,0 +1,267 @@
...
+            protocol: if tls.uses_tls() {
+                "https".to_string()
+            } else {
+                "http".to_string()
+            },
+        }
+    }
+
+    pub fn oauth2_token(&self) -> String {
+        format!(
```
sbernauer, line 160: nit: We could use https://docs.rs/url/latest/url/struct.Url.html for all URL-related computation and return an URL
Watch out, https://docs.rs/url/latest/url/struct.Url.html#method.join sucks :D

sbernauer (commented):
Only minor comments.
Did not test it, you mentioned some Python script for simulating Entra?

rust/user-info-fetcher/src/backend/entra.rs
maltesander: That is probably because the whole thing was copy pastad from the keycloak impl :)

deploy/helm/opa-operator/crds/crds.yaml
```diff
@@ -162,54 +162,44 @@ spec:
                                   description: The Microsoft Entra tenant ID.
                                   type: string
                                 tls:
-                                  default:
```
sbernauer: I think it actually makes sense to default to webPki instead of null. I don't know what changes you did in Rust that resulted in this CRD change

rust/operator-binary/src/crd/user_info_fetcher.rs
```diff
@@ -110,6 +117,34 @@ pub mod versioned {
         pub additional_group_attribute_filters: BTreeMap<String, String>,
     }
 
+    #[derive(Clone, Debug, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct EntraBackend {
+        /// Hostname of the token provider, defaults to `login.microsoft.com`.
+        #[serde(default = "entra_default_host_token")]
+        pub hostname_token: HostName,
```
sbernauer, line 125: nit: I think I would slightly prefer `token_provider_hostname`

rust/operator-binary/src/crd/user_info_fetcher.rs
```diff
@@ -110,6 +117,34 @@ pub mod versioned {
...
+
+        /// Hostname of the user info provider, defaults to `graph.microsoft.com`.
+        #[serde(default = "entra_default_host_graph")]
+        pub hostname_graph: HostName,
```
sbernauer, line 129: nit: I think I would slightly prefer `graph_hostname` or `user_info_hostname`

rust/operator-binary/src/crd/user_info_fetcher.rs
```diff
@@ -110,6 +117,34 @@ pub mod versioned {
...
+
+        /// Port of the identity provider. Defaults to 443.
+        #[serde(default = "entra_default_port")]
+        pub port: u16,
+
+        /// The Microsoft Entra tenant ID.
+        pub tenant_id: String,
+
+        /// Use a TLS connection. Should usually be set to WebPki.
+        #[serde(default = "default_tls_web_pki")]
```
sbernauer, line 139: Can you please explain as a comment here why we can not serde(flatten) TlcClientDetails?

rust/user-info-fetcher/src/backend/entra.rs
```diff
@@ -0,0 +1,295 @@
...
+
+use crate::{Credentials, UserInfo, UserInfoRequest, http_error, utils::http::send_json_request};
+
+#[derive(Snafu, Debug)]
+pub enum Error {
+    #[snafu(display("failed to get access_token"))]
+    AccessToken { source: crate::utils::http::Error },
+
+    #[snafu(display("failed to search for user"))]
+    SearchForUser { source: crate::utils::http::Error },
```
sbernauer, line 17: Can you please include the information what user we are searching for?

rust/user-info-fetcher/src/backend/entra.rs
```diff
@@ -0,0 +1,295 @@
...
+
+    let user_info = match req {
+        UserInfoRequest::UserInfoRequestById(req) => {
+            let user_id = req.id.clone();
+            send_json_request::<UserMetadata>(
+                http.get(entra_endpoint.user_info(&user_id))
+                    .bearer_auth(&authn.access_token),
+            )
+            .await
+            .context(UserNotFoundByIdSnafu { user_id })?
```
sbernauer, line 121: Just for the record, I think the error message is a bit misleading. It can also be a network error or permission problem.
No need to change though

rust/user-info-fetcher/src/backend/entra.rs
```diff
@@ -0,0 +1,295 @@
...
+            .context(SearchForUserSnafu)?
+        }
+    };
+
+    let groups = send_json_request::<GroupMembershipResponse>(
+        http.get(entra_endpoint.group_info(&user_info.id))
+            .bearer_auth(&authn.access_token),
+    )
+    .await
+    .context(RequestUserGroupsSnafu {
```
sbernauer, line 139: nit: Please use `with_context` to avoid needless cloning

sbernauer (changes requested):

rust/user-info-fetcher/src/backend/entra.rs
sbernauer: Thanks!

sbernauer (changes requested):
Thanks for addressing the review!

rust/operator-binary/src/crd/user_info_fetcher.rs
```diff
@@ -110,6 +117,40 @@ pub mod versioned {
         pub additional_group_attribute_filters: BTreeMap<String, String>,
     }
 
+    #[derive(Clone, Debug, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct EntraBackend {
+        /// Hostname of the token provider, defaults to `login.microsoft.com`.
+        #[serde(default = "entra_default_token_endpoint")]
+        pub token_endpoint: HostName,
```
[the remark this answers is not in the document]
maltesander: done https://github.com/stackabletech/opa-operator/pull/712/commits/b1913ca28f010aa12e846c6571d6f5a630283729

rust/user-info-fetcher/src/backend/entra.rs
```diff
@@ -0,0 +1,315 @@
...
+            })?;
+
+        Ok(Self {
+            token_endpoint_url,
+            user_info_endpoint_url,
+        })
+    }
+
+    pub fn oauth2_token(&self) -> String {
+        self.token_endpoint_url.to_string()
```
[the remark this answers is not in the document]
maltesander: done https://github.com/stackabletech/opa-operator/pull/712/commits/8c07822b12e180ed8b13abfc25bd0996d0c6eee7

sbernauer (commented):
Only one CRD change question left.
As this is a CRD change I think we should put it on the decisions board (even for experimental features)

deploy/helm/opa-operator/crds/crds.yaml
```diff
@@ -137,6 +139,81 @@ spec:
...
+                              properties:
+                                clientCredentialsSecret:
+                                  description: |-
+                                    Name of a Secret that contains client credentials of an Entra account with permissions `User.ReadAll` and `GroupMemberShip.ReadAll`.
+
+                                    Must contain the fields `clientId` and `clientSecret`.
+                                  type: string
+                                port:
+                                  default: 443
+                                  description: Port of the identity provider. Defaults to 443.
```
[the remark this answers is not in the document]
maltesander: Hmm the decision is based on tls https://github.com/stackabletech/opa-operator/blob/616b51463411d0baacff6f797c8e57b1efa54d57/rust/user-info-fetcher/src/backend/entra.rs#L173
I think thats the only thing it should be based on. You could argue not handing over the port in the endpoints here https://github.com/stackabletech/opa-operator/blob/616b51463411d0baacff6f797c8e57b1efa54d57/rust/user-info-fetcher/src/backend/entra.rs#L176 and let Url do its thing? 
I dont see the point in using an option when i dont have to?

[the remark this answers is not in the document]
maltesander: https://github.com/stackabletech/opa-operator/pull/712/commits/4d10eaaaf8d0e83ab31a41d580335b087c89e082 as discussed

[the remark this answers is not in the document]
sbernauer: Many thanks!

sbernauer (approved):
LGTM, but I think this needs to go through the decision board

sbernauer (approved):

lfrancke:
Could you add a link to the docs and release notes please?

maltesander:
Docs: https://docs.stackable.tech/home/nightly/opa/usage-guide/user-info-fetcher/#backend-entra
Release Snippet: ```The user-info-fetcher backend now supports fetching user groups from Microsoft Entra experimentally.```

PR #745, opened 2025-07-17
fix: Add RBAC permission to patch events

sbernauer:
Needed since https://github.com/stackabletech/operator-rs/pull/938

Not 100% sure why the product needs this, but it was this way before.
Also, fix a wrongly placed "patch"

### DIFF
--- a/deploy/helm/opa-operator/templates/roles.yaml
+++ b/deploy/helm/opa-operator/templates/roles.yaml
@@ -82,7 +82,6 @@ rules:
       - events.k8s.io
     resources:
       - events
-      - patch
     verbs:
       - create
       - patch
@@ -132,6 +131,7 @@ rules:
       - events.k8s.io
     resources:
       - events
+      - patch
     verbs:
       - create
 {{ if .Capabilities.APIVersions.Has "security.openshift.io/v1" }}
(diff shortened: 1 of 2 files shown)


deploy/helm/opa-operator/templates/roles.yaml
```diff
@@ -132,6 +131,7 @@ rules:
       - events.k8s.io
     resources:
       - events
+      - patch
```
NickLarsenNZ, line 134: This should be the verb, right?

deploy/helm/opa-operator/templates/roles.yaml
```diff
@@ -132,6 +131,7 @@ rules:
...
     verbs:
       - create
```
sbernauer, line 136:
```suggestion
      - create
      - patch
```

deploy/helm/opa-operator/templates/roles.yaml
sbernauer: Yep, you are right, too many PRs

deploy/helm/opa-operator/templates/roles.yaml
```diff
@@ -132,8 +131,10 @@ rules:
       - events.k8s.io
     resources:
       - events
+      - patch
```
sbernauer, line 134:
```suggestion
```

NickLarsenNZ (changes requested):
Missing changelog

lfrancke:
The Arch packages have been upgraded: https://repology.org/project/stackablectl/versions

PR #771, opened 2025-10-27
feat: Add EoS checker

Techassi:
Part of https://github.com/stackabletech/issues/issues/733. This adds the EoS checker to the operator.


deploy/helm/opa-operator/crds/crds.yaml
```diff
@@ -459,6 +507,7 @@ spec:
                                           - ERROR
                                           - FATAL
                                           - NONE
+                                          - null
```
sbernauer: Just FYI that this is introduced by kube 2. I *think* it doesn't hurt to have it here

PR #796, opened 2026-01-14
chore: Switch from `rustls-pemfile` to `rustls-pki-types` to resolve RUSTSEC-2025-0134

sbernauer:
Fixes https://rustsec.org/advisories/RUSTSEC-2025-0134 that popped up in https://github.com/stackabletech/opa-operator/pull/795


NickLarsenNZ (approved):
LGTM

It should be noted that a crate was swapped out (`rustls-pemfile` -> `rustls-pki-types`).

Issue #791, opened 2025-12-10
Stabilise Entra backend for User Info Fetcher


lfrancke:
In [25.7](https://github.com/stackabletech/opa-operator/pull/712) we added a backend for the UIF and marked it as experimental.

It's been in use in production for a while and we believe we can mark it as stable now.

PR #801, opened 2026-01-23
feat(crd): Stabilise Entra backend

Techassi:
Implements #791.

This PR stabilises the Entra backend by introducing `v1alpha2` and adds a conversion webhook to convert between `v1alpha1` and `v1alpha2`.

There is no Entra specific integration test with which this change can be tested, but applying a `v1alpha1` OpaCluster results in a successful conversion to `v1alpha2`. The downgrade also works without any issues.

_Note for reviewers: It might be easier to review this on a commit-by-commit basis._


Techassi:
Oh yeah, you are right. I will update them asap.

PR #813, opened 2026-03-11
fix(docs): Update getting-started script pre-26.3.0

Techassi:
## Check and Update Getting Started Script

Part of <https://github.com/stackabletech/issues/issues/826>

This getting-started script was technically fine as is (it ran through without any issues), but I still opted to include both waiting conditions, because it generally a good thing to do going forward.


sbernauer (approved):
Quickly grepped for other 1.8.0 versions, LGTM

Issue #848, opened 2026-07-02
Resource Info Fetcher (RIF)


Maleware:
## Summary
Design and implement a **Resource Info Fetcher** (RIF), a sidecar service that supplies OPA with metadata about *resources* (datasets, tables, topics, charts, …). It is supposed to be the resource-side part of **User Info Fetcher**.

- UIF: *"which groups is this user in?"*
- RIF: *"what do we know about this resource?"*.

Initially we want to implement DataHub as backend and then might include more as we progress. Designing this we want to provide the option to add more backends non-braking as well as think about a more generic approach once we have experience with different backends.

## Background
- OPA policies usefulness profits from attributes attached to the resource being accessed (owner, sensitivity tags, glossary terms, data product, …).
- Uniform Resource Name ([URN](https://datatracker.ietf.org/doc/html/rfc8141?__cf_chl_f_tk=TBPrlqtDrlNe5V6xFm52kI6Idtp23AK7Ib7sBb11sSs-1782989757-1.0.1.1-_RtQ2k1wbclrmvbvscFwZ4hPExP.Fv_iXcjc6oOJ1aQ)): `urn:<NAMESPACE-IDENTIFIER>:<NAMESPACE-SPECIFIC-STRING>`
- Commonly metadata lives in catalog systems (DataHub, OpenMetadata, …).
- The UIF pattern has proven that a sidecar with a stable API + swappable backends is a good fit for this problem shape.
- Spike by @soenkeliebau:
  - branch comparison: <https://github.com/stackabletech/opa-operator/compare/main...feat/resource-info-fetcher>
  - prototype repo: <https://github.com/stackabletech/resourceinfofetcher>

## Goals
- HTTP API served by RIF, consumed by OPA (via `http.send` from Rego, analogous to UIF).
- Backend adapters for DataHub (at first).
- Cache layer between RIF and the backend.
- Ship as a **separate binary** from UIF (working assumption — revisit later).

## Non-goals

- Write-back or metadata authoring — RIF is read-only.

## Out of scope
- Come up with an abstraction layer shared by all RIF backends. For v1 every backend returns it's own "proprietary" response
-  OpenMetadata, Synabi D-QUANTUM, Collibra

## Data model (starting point from the spike, can be used or discraded)

Two structs live in the prototype repo:

```rust
pub struct ResourceInfo {
    #[serde(default)] pub tags: Vec<String>,
    #[serde(default)] pub glossary_terms: Vec<String>,
[...]

PR #863, opened 2026-07-21
feat: Add resource-info-fetcher

sbernauer:
## Description

Part of https://github.com/stackabletech/opa-operator/issues/848

Yes, the diff looks big, but most of it are lockfiles and kuttl tests

## CRD change

Basically the same as for the user-info-fetcher. The only difference is the `env` field, which is DataHub specific.

Look at `extra/crds.yaml` for the concrete CRD change

```yaml
  clusterConfig:
    resourceInfo: # <1>
      backend:
        dataHub:
          hostname: datahub-gms.my-namespace.svc.cluster.local
          port: 8080
          rootPath: / # optional, defaults to /
          tls:
            verification:
              server:
                caCert:
                  secretClass: tls # <2>
          credentialsSecretName: resource-info-fetcher-credentials # <3>
          env: PROD # <4>
      cache: # optional, enabled by default
        maxEntries: 42 # optional, defaults to no limit
        entryTimeToLive: 60s # optional, defaults to 60s
---
apiVersion: v1
kind: Secret
metadata:
  name: resource-info-fetcher-credentials
stringData:
  token: <datahub-personal-access-token> # <3>
```

## API of RIF

The rego rules and HTTP API of RIF: https://github.com/stackabletech/opa-operator/blob/spike/rif/docs/modules/opa/pages/usage-guide/resource-info-fetcher.adoc

### Author

- [x] Changes are OpenShift compatible
- [ ] CRD changes approved
- [x] CRD documentation for all fields, following the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).
- [x] Helm chart can be installed and deployed operator works
- [x] Integration tests passed (for non trivial changes)
- [x] Changes need to be "offline" compatible
- [x] Links to generated (nightly) docs added
- [x] Release note snippet added

### Reviewer

- [x] Code contains useful logging statements
- [x] (Integration-)Test cases added
- [x] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).
- [x] Changelog updated

### Acceptance

##### Added

- Add an initial version of resource-info-fetcher, which is similar to user-info-fetcher, but allows to fetch additional metadata about resource information from a data catalog.
  For now only DataHub is supported.
  Also, a rego-rule library has been added to make it easier to call resource-info-fetcher from within OPA.
  The API (especially the response) might change in the future once more data catalogs are supported

### Nightly docs

As soon as this PR is merged the docs will show up at https://docs.stackable.tech/home/nightly/opa/usage-guide/resource-info-fetcher/


lfrancke:
Because we had it yesterday with OpenLineage: I believe it could make sense to add a "path" or "endpoint" field. Just in case DataHub is deployed behind some proxy. <host>:<port>/<path>/<here is the normal datahub stuff>

sbernauer:
> I believe it could make sense to add a "path" or "endpoint" field

Good point! For consistency reasons with UIF and AuthenticationClass I went with `rootPath` in 215d4b9. Feel free to have a look

rust/resource-info-fetcher/src/api.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub id: u64,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Deserialize)]
+pub struct RawIdentifier {
+    pub identifier: String,
+}
+
+/// Generates the trivial `From<Params> for ResourceInfoRequest` conversions, so each HTTP handler
+/// can turn its deserialized query parameters into a [`ResourceInfoRequest`] via `.into()`. Adding a
```
Maleware, line 162:
```suggestion
/// can turn its deserialized query parameters into a [`ResourceInfoRequest`] via `.from()`. Adding a
```

rust/resource-info-fetcher/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+    config: PathBuf,
+
+    #[clap(long, env)]
+    credentials_dir: PathBuf,
+}
+
+#[derive(Clone)]
+struct AppState {
+    backend: Arc<ResolvedBackend>,
+    // Note: Although we might no talk JSON to the underlying backend, we always return JSON as a
```
Maleware:
```suggestion
    // Note: Although we might not talk JSON to the underlying backend, we always return JSON as a
```

Maleware (commented):
Just two typos, thanks you! Test passes locally, thus LTGTM otherwise.

rust/resource-info-fetcher/src/api.rs
sbernauer: https://doc.rust-lang.org/std/convert/trait.From.html
From automatically implements Into, so the caller can pick.
In https://github.com/stackabletech/opa-operator/blob/970da87fb700ef07b70e1a4f5dfbaee69b876d1a/rust/resource-info-fetcher/src/main.rs#L185 we actually use `Into<ResourceInfoRequest>`. In general I feel in this situation a into feels more natural than a from

sbernauer:
Please vote on the CRD change as well as the rego rule API on this comment. Feel free to also look at the JSON API, but I'd say that's an implementation detail we are allowed to change in the future

rust/resource-info-fetcher/src/backend/data_hub/mod.rs
```diff
@@ -0,0 +1,378 @@
...
+                path: token_path.display().to_string(),
+            })?
+            .trim()
+            .to_owned();
+
+        let mut client_builder = ClientBuilder::new();
+        client_builder = utils::tls::configure_reqwest(&config.tls, client_builder)
+            .await
+            .context(ConfigureTlsSnafu)?;
+        let http_client = client_builder.build().context(ConstructHttpClientSnafu)?;
```
maltesander: This uses request client builder and does not have any timeouts set.
Check the UIF backends and please use `utils::http::client_builder()` instead.

rust/resource-info-fetcher/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+
+    let config: v1alpha1::Config = read_config_file(&args.config)
+        .with_context(|_| ParseConfigFileSnafu { path: args.config })?;
+    let backend = Arc::new(resolve_backend(config.backend, &args.credentials_dir).await?);
+    let resource_info_cache = {
+        Cache::builder()
+            .name("resource-info")
+            .time_to_live(*config.cache.entry_time_to_live)
+            .build()
+    };
```
maltesander: The sidecar has 128MB memory and this cache can grow forever. UIF is limited/keyed on usernames from a directory, so this is limited. Set a max capacity (or make it configurable)?

rust/resource-info-fetcher/src/backend/data_hub/mod.rs
```diff
@@ -0,0 +1,378 @@
...
+    /// Resolves a DataHub backend by loading credentials from the filesystem.
+    #[instrument(skip_all)]
+    pub async fn resolve(
+        config: v1alpha1::DataHubBackend,
+        credentials_dir: &Path,
+    ) -> Result<Self, Error> {
+        let token_path = credentials_dir.join("token");
+
+        // Trim trailing whitespace/newlines so the value is safe to use in an HTTP header.
+        let token = tokio::fs::read_to_string(&token_path)
```
maltesander, line 206: This is never refreshed? Will go stale and produce 401 until restart?

rust/resource-info-fetcher/src/backend/data_hub/graphql.rs
```diff
@@ -0,0 +1,376 @@
...
+const RESOURCE_INFO_QUERY: &str = r#"
+query ResourceInfo($urn: String!) {
+  entity(urn: $urn) {
+    ...ResourceInfo
+  }
+}
+fragment ResourceInfo on Entity {
+  # DataHub has no direct "dataProduct" field on assets; membership is a graph edge that points from
+  # the data product to its assets. From the asset's side it is therefore an INCOMING relationship.
+  dataProducts: relationships(input: {types: ["DataProductContains"], direction: INCOMING, count: 10}) {
```
maltesander: What if there is something belonging to 11 data products? Why limit to 10? This will hard fail policies that should succeed. Can we paginate that or warn (if a much higher limit than 10 (e.g. 1000)) is reached somehow?

docs/modules/opa/pages/usage-guide/resource-info-fetcher.adoc
```diff
@@ -0,0 +1,179 @@
+= Resource info fetcher
```
maltesander, line 1: Not referenced in nav.adoc? 
```
xref:opa:usage-guide/resource-info-fetcher.adoc[]
```

rust/resource-info-fetcher/src/api.rs
```diff
@@ -0,0 +1,143 @@
...
+/// deserializes its query parameters into the variant's payload struct and the backend then maps
+/// the request to whatever the concrete backend needs (for DataHub: a URN, see [`urn_for_request`]).
+///
+/// Each variant wraps its own parameter struct rather than inlining the fields, so a single struct
+/// serves as both the HTTP query-parameter target ([`axum::extract::Query`]) and the enum payload —
+/// there is no second copy of the field list to keep in sync.
+///
+/// [`urn_for_request`]: crate::backend::data_hub::resource_to_urn_mapping::urn_for_request
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum ResourceInfoRequest {
```
maltesander, line 31: The `instance` is always required. Datahub (defaults) leave that out and a trino table will become e.g. `urn:li:dataset:(urn:li:dataPlatform:trino,tpch.sf1.customer,PROD)` and the `tableResourceInfo` always builds `{instance}.tpch.sf1.customer` which will be empty and therefore fail a correct rego?

Can we make instance optional and skip?

maltesander (changes requested):
First round, did not check docs etc. properly yet and not testing yet.

docs/modules/opa/pages/usage-guide/resource-info-fetcher.adoc
sbernauer: Good catch, bf85e53

rust/resource-info-fetcher/src/backend/data_hub/graphql.rs
sbernauer: That is a real problem I also noticed but forgot about. It should never happen in reality, so I added a snafu error for it, so it loudly fails and does not silently influence authorization: cd9a38ed0e59bdbb766065ee9f914f0282145a05

rust/resource-info-fetcher/src/backend/data_hub/mod.rs
sbernauer: Good catch, fixed in c20d1d1

rust/resource-info-fetcher/src/main.rs
sbernauer: I copied this from UIF and was also surprised there is no cache limit.
However, I disagree that UIF is really different, as the number of users in Keycloak (I guess) is similarly unlimited than the number of resources in DataHub.
So I would be in favor of adding the limit to both caches.
Did so in f481d44

rust/resource-info-fetcher/src/backend/data_hub/mod.rs
sbernauer: Yes, but `ResolvedKeycloakBackend::resolve`, `ResolvedEntraBackend::resolve`, `ResolvedOpenLdapBackend::resolve` and maybe others do the exact same thing.
According to https://docs.datahub.com/docs/authentication/personal-access-tokens you can create a PAT without expiration.
I see that is not 100% ideal, but also IMHO the status quo basically everywhere else in the platform as well

rust/resource-info-fetcher/src/main.rs
maltesander: True and +1 for UIF fix. RIF can just grow much quicker in size per request, but yeah the problem is there for both. Thanks!

rust/resource-info-fetcher/src/api.rs
sbernauer: I was thinking of this as well (I actually started without the instance).
I think we should very strongly recommend users to set the instance when ingesting data into DataHub.
Also we likely want to support other backends in the future.
If they don't have such a strange concept of "there is only one Trino" they will require us to send a instance for every request.
So I fear we would have a problem with future backends if we make it optional.

rust/resource-info-fetcher/src/backend/data_hub/mod.rs
maltesander: No i dont think so. They read the client credentials, not the token? https://github.com/stackabletech/opa-operator/blob/8402d626e3c9bf8f88e89f5478c6492f5d9ff8e1/rust/user-info-fetcher/src/backend/keycloak.rs#L189 (same for entra).
This is not ideal either, we should just cache it and refresh on 401.

rust/resource-info-fetcher/src/api.rs
maltesander: Hmm i dont think we should *strongly* recommend how people use a non stackable product?

I think the scope of this PR is datahub, so i would try to represent everything datahub can do with reasonable abstraction? Removing the optional part can then be done once another backend lands?

I feel this is a too generalized and therefore cutting into datahub funtionality?

rust/resource-info-fetcher/src/backend/data_hub/mod.rs
```diff
@@ -0,0 +1,398 @@
...
+
+impl http_error::Error for Error {
+    fn status_code(&self) -> StatusCode {
+        match self {
+            Self::ReadToken { .. } => StatusCode::SERVICE_UNAVAILABLE,
+            Self::ConfigureTls { .. } => StatusCode::SERVICE_UNAVAILABLE,
+            Self::ConstructHttpClient { .. } => StatusCode::SERVICE_UNAVAILABLE,
+            Self::BuildDataHubEndpoint { .. } => StatusCode::BAD_REQUEST,
+            Self::ExecuteGraphQlQuery { .. } => StatusCode::INTERNAL_SERVER_ERROR,
+            Self::GraphQlErrors { .. } => StatusCode::INTERNAL_SERVER_ERROR,
```
maltesander: I think this is user-controlled, e.g. a bad identifier is rather a client error?
```suggestion
            Self::GraphQlErrors { .. } => StatusCode::BAD_REQUEST,
```

maltesander (commented):
Claude reproduced a bigger security issue (policies can pass if backend is gone).

## Silent fail-open when the backend is unavailable

| Backend state | `allow if tag==public` | `deny if tag==pii` |
|---|---|---|
| healthy | deny | **deny** ✅ |
| PAT rejected (401) | deny | **allow** ❌ |
| GMS scaled to 0 | deny | **allow** ❌ |

All three containers report `ready=true` throughout. There is no readiness probe on the sidecar.

- **Repro:** `kubectl -n $NS scale deploy datahub-datahub-gms --replicas=0`, wait 60s for the cache to drain, query any policy that excludes a tag.
- **Fix:** add a readiness probe in `controller/build/resource/daemonset/resource_info_fetcher.rs`

## Tags and domains are not inherited from the parent container

Tagging schema `tpch.sf1` with `pii` leaves child table `nation` reporting `tags:[public], domain:null`. A `deny if pii` policy allows it. DataHub exposes the parent in the *same* query (`container { tags domain }`), so this is one field away.

- **Repro:** `addTag`/`setDomain` on the schema container URN, wait out the 60 s TTL, query the table.
- **Fix:** add `container { tags {...Tags} domain {...Domain} }` to the fragment at `graphql.rs:30`, or state the non-inheritance explicitly in the docs.

rust/resource-info-fetcher/src/api.rs
Maleware: Ah okay allright. I was looking at the function beneath which reads `from` thought that might want to match.

rust/operator-binary/src/controller/build/resource/daemonset/resource_info_fetcher.rs
```diff
@@ -0,0 +1,105 @@
...
+};
+
+use crate::controller::{
+    ValidatedCluster, ValidatedOpaConfig,
+    build::{
+        self,
+        resource::daemonset::{
+            CONFIG_DIR, CONFIG_VOLUME_NAME, RESOURCE_INFO_FETCHER_CREDENTIALS_DIR,
+            RESOURCE_INFO_FETCHER_CREDENTIALS_VOLUME_NAME, add_stackable_rust_cli_env_vars,
+            container_name, sidecar_container_log_level, sidecar_resource_requirements,
```
Maleware: Looking at the bundle-builder, we set a mount volume for logs being collected by vector.

Neither UIF or RIF have this volume mounted and thus logs cannot be collected by vector.

Those volumes should be mounted into RIF and UIF to fix that. Additionally, the integration test for logging should test for this.

Maleware (changes requested):

rust/resource-info-fetcher/src/api.rs
sbernauer: > Hmm i dont think we should strongly recommend how people use a non stackable product?

Ok for me, happy to remove the "strongly" :P Still a stupid idea ^^

> I feel this is a too generalized and therefore cutting into datahub funtionality?

Currently the *request* is abstracted away from DataHub (the *response* is DH-specifc). If we make the `instance` optional we loose that, which I find sad ^^

If we go with mandatory for now, we can always make it optional later if a customer requests it. Going the other way is a breaking change.

maltesander:
```
--- PASS: kuttl (549.95s)
    --- PASS: kuttl/harness (0.00s)
        --- PASS: kuttl/harness/smoke_opa-1.12.3_openshift-false_use-tls-true (56.38s)
        --- PASS: kuttl/harness/resources_opa-latest-1.16.2_openshift-false (20.21s)
        --- PASS: kuttl/harness/logging_opa-1.16.2_openshift-false (61.86s)
        --- PASS: kuttl/harness/logging_opa-1.12.3_openshift-false (61.94s)
        --- PASS: kuttl/harness/smoke_opa-1.16.2_openshift-false_use-tls-true (56.42s)
        --- PASS: kuttl/harness/smoke_opa-1.16.2_openshift-false_use-tls-false (52.11s)
        --- PASS: kuttl/harness/cluster-operation_opa-latest-1.16.2_openshift-false (33.07s)
        --- PASS: kuttl/harness/smoke_opa-1.12.3_openshift-false_use-tls-false (52.29s)
        --- PASS: kuttl/harness/data-hub-resource-info_opa-latest-1.16.2_trino-latest-481_kafka-latest-4.2.1_superset-latest-6.1.0_data-hub-1.0.3_data-hub-prerequisites-0.3.0_openshift-false (398.77s)
        --- PASS: kuttl/harness/config-overrides_opa-latest-1.16.2_openshift-false (49.57s)
        --- PASS: kuttl/harness/openldap-user-info_opa-latest-1.16.2_openshift-false (62.96s)
        --- PASS: kuttl/harness/aas-user-info_opa-latest-1.16.2_openshift-false (52.12s)
        --- PASS: kuttl/harness/keycloak-user-info_opa-latest-1.16.2_keycloak-23.0.1_openshift-false (92.70s)
PASS
```

rust/resource-info-fetcher/src/backend/data_hub/graphql.rs
```diff
@@ -0,0 +1,833 @@
...
+
+  # DataHub has no direct "dataProduct" field on assets; membership is a graph edge that points from
+  # the data product to its assets. From the asset's side it is therefore an INCOMING relationship.
+  # `total` is the number of edges DataHub has, which we compare against the number we received to
+  # detect that the page size was too small.
+  dataProducts: relationships(input: {types: ["DataProductContains"], direction: INCOMING, count: $dataProductsCount}) {
+    total
+    relationships { ...DataProduct }
+  }
+  ... on Dataset   { tags { ...Tags } ownership { ...Owners } domain { ...Domain } }
```
Maleware: I became aware during testing, that we currently cannot differentiate between metadata empty and resource does not exists. 

This is because we only check for
```
        let Some(entity) = data.and_then(|data| data.entity) else {
            debug!(%urn, "DataHub returned no entity; responding with empty resource info");
            return Ok(graphql::Entity::default());
        };
```
which returns a empty set for both cases, not found and empty. This is in turn because Entity::into_response() takes tags and if empty returns none, which will be true for both cases.

I looked up datahub docs and they offer a bool [exists](https://docs.datahub.com/docs/graphql/queries#entityexists) which would, added to the graphQL query, make those cases distinguishable. 

Thus I'd suggest something along the lines:
```
... on Dataset   { exists tags { ...Tags } ownership { ...Owners } domain { ...Domain } }
```
and rather probe for it in datahub::mod.rs e.g. like:
```
if entity.exists == Some(false) {
    debug!(%urn, "DataHub does not hold this entity; reporting it as not found");
    return NotFoundSnafu { urn: urn.clone() }.fail();
}
```

I'd argue a resource not found should error out to make the user aware of it. If we don't, no metadata present becomes equal to resource not found and might end in a debugging rabbit hole.

rust/resource-info-fetcher/src/backend/data_hub/mod.rs
```diff
@@ -0,0 +1,496 @@
...
+impl ResolvedDataHubBackend {
+    /// Resolves a DataHub backend by loading credentials from the filesystem.
+    #[instrument(skip_all)]
+    pub async fn resolve(
+        config: v1alpha1::DataHubBackend,
+        credentials_dir: &Path,
+    ) -> Result<Self, ResolveError> {
+        let token_path = credentials_dir.join("token");
+
+        // Trim trailing whitespace/newlines so the value is safe to use in an HTTP header.
```
Maleware: This should go into info-fetcher-commons and be used in user-info-fetcher as well (both backends entra and keycloak).

It doesn't trim whitespaces in credentials at this moment, but I think it should.

rust/info-fetcher-commons/src/utils/token.rs
```diff
@@ -0,0 +1,382 @@
...
+///
+/// A token is minted, then travels to the backend and is validated there, so handing out one that is
+/// about to expire risks it being rejected mid-request. Refreshing slightly early avoids that without
+/// needing to know anything about the backend's clock.
+pub const EXPIRY_MARGIN: Duration = Duration::from_secs(30);
+
+/// A freshly minted bearer token.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MintedToken {
+    pub token: String,
```
Maleware: Nit: It's not a problem right now, since the token never get's printed in a logging statement, but I'd argue that it can happen easily by accident.

I think it might be better to make this struct denying to print itself in logging statements by default so it must be a active decision to make it printable in the future.

rust/resource-info-fetcher/src/cache.rs
```diff
@@ -0,0 +1,191 @@
...
+/// than this, it wins, because moka evicts at the earliest of the two.
+const FAILURE_TIME_TO_LIVE: Duration = Duration::from_secs(5);
+
+/// What a lookup produced, as held in the cache.
+///
+/// Failures are cached as well as successes, so the variants share one key space and one capacity
+/// limit, and so that moka coalesces concurrent lookups of a failing key just as it does a
+/// successful one.
+#[derive(Clone)]
+pub enum CachedResponse {
```
Maleware: I like this idea. I think this should also go into info-fetcher-commons and should be used in user-info-fetcher as well.

(If I understood UIF correctly, that struct there cannot hold the failure case. At https://github.com/stackabletech/opa-operator/blob/spike/rif/rust/user-info-fetcher/src/main.rs#L191

Maleware (changes requested):

rust/resource-info-fetcher/src/backend/data_hub/mod.rs
maltesander: done https://github.com/stackabletech/opa-operator/pull/863/commits/976fbd394117b3bcaf58aa96b4dd5ce51dd50162

rust/info-fetcher-commons/src/utils/token.rs
maltesander: done https://github.com/stackabletech/opa-operator/pull/863/commits/748a69cb8e30145a2a86faa3119bdd1db5d5e195

rust/resource-info-fetcher/src/cache.rs
maltesander: done https://github.com/stackabletech/opa-operator/pull/863/commits/1a96050136c305fc1c2bf847131284753db26ea9

rust/resource-info-fetcher/src/backend/data_hub/graphql.rs
maltesander: We talked about that and it is documented in `usage-guide/resource-info-fetcher.adoc`:

Behaviour when metadata is unavailable: the resource fetcher cannot know whether an empty record should mean allow or deny for a given policy...

So i would keep this as is.

rust/resource-info-fetcher/src/backend/data_hub/mod.rs
```diff
@@ -0,0 +1,487 @@
...
+    #[instrument(skip(self))]
+    async fn query_entity(&self, urn: &Urn) -> Result<graphql::Entity, Error> {
+        trace!(%urn, %self.graphql_url, "Sending GraphQL query to DataHub");
+
+        let response: graphql::GraphQlResponse = send_json_request(
+            self.http_client
+                .post(self.graphql_url.clone())
+                // Authenticate with a DataHub Personal Access Token (a bearer JWT). DataHub's
+                // Metadata Service Authentication verifies it and resolves it to the token's actor.
+                .bearer_auth(self.token.expose())
```
Maleware: This token only get read once at start up. Means rotating a expired token would only work if OPA pods get restarted.

IMO this should have a mechanism to reload this token without the need of an pod restart, or if not feasable, we should auto rolling restart to ensure tokens get distributed across all OPA pods without a user interacting with it.

maltesander: The PAT is userdefined, so updating the secret will rotate it.
Edit: No it doesnt, we dont watch secrets or automatically restart. Needs investigation.

rust/resource-info-fetcher/src/backend/data_hub/graphql.rs
maltesander: Used a combination of `exists` and `status.removed` to differentiate between (non-) existing resources and resources without metadata: https://github.com/stackabletech/opa-operator/pull/863/commits/aaa18c771e4ef488bd2c90c9a7899e3e3b7a088b

maltesander (approved):
LGTM!

Maleware (approved):
Looked through the code and I've tested the new functionalities and they seem to work as expected. LGTM.

Thanks! :)

Maleware:
### Release notes

##### Added

- Add an initial version of resource-info-fetcher, which is similar to user-info-fetcher, but allows to fetch additional metadata about resource information from a data catalog.
  For now only DataHub is supported.
  Also, a rego-rule library has been added to make it easier to call resource-info-fetcher from within OPA.
  The API (especially the response) might change in the future once more data catalogs are supported

lfrancke:
Why is this marked as "action-required"?

Issue #444, opened 2023-05-12
Automatically migrate container name 23.1 -> 23.4


sbernauer:
### Affected version

23.4.0

### Current and expected behavior

In 23.1 the bundle-builder container within the OPA DaemonSet was called `opa-bundle-builder`.
In https://github.com/stackabletech/opa-operator/pull/420 it was renamed to `bundle-builder`.
When upgrading a OPA cluster 23.1 -> 23.4 the opa-operator *patches* the DaemonSet, so we end up with both containers.
The `bundle-builder` containers fails with
`thread 'main' panicked at 'error binding to 0.0.0.0:3030: error creating server listener: Address already in use (os error 98)', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/warp-0
.3.3/src/server.rs:213:27`

### Would you like to work on fixing this bug?

yes

PR #872, opened 2026-08-07
refactor: Extract apply and update status steps

maltesander:
## Description

- extracts an apply step (`controller/apply.rs`) with an `Applier`, which owns the `ClusterResources`, applies every resource kind and deletes orphaned resources
- extracts an update status step (`controller/update_status.rs`), which computes the `OpaClusterStatus` from the applied resources
- adds a type state marker to `KubernetesResources` (`Prepared`, `Applied`), so `update_status` can only be called with resources that were actually applied, not merely built. `Applier::apply` destructures `KubernetesResources` without a rest pattern, so a new resource kind fails to compile instead of silently never being applied.
- moves the legacy "opacluster" field manager patch (#444) into a private method on the `Applier`, running against the applied DaemonSets. Only the name and namespace are read from those objects, so this is equivalent to the previous behaviour.

### Author

- [ ] CRD changes approved
- [ ] CRD documentation for all fields, following the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).
- [ ] Changes need to be "offline" compatible

### Reviewer

- [x] Code contains useful logging statements
- [x] Changelog updated

### Acceptance

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@ All notable changes to this project will be documented in this file.
 - The reconciler now applies resources and derives the cluster status in discrete
   apply and update_status steps for the `opa_controller` ([#872]).
 
-
 [#852]: https://github.com/stackabletech/opa-operator/pull/852
 [#861]: https://github.com/stackabletech/opa-operator/pull/861
 [#867]: https://github.com/stackabletech/opa-operator/pull/867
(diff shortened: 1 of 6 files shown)


siegfriedweber (dismissed):
LGTM

siegfriedweber (approved):

PR #873, opened 2026-08-11
Feature: Allow workloadKind: Deployment

Maleware:
## Description

~~HINT: Takes `spike/rif` as base as changes overlap. Please merge https://github.com/stackabletech/opa-operator/pull/863 first.~~

Allow opa being deployed as workload deployment.

This PR contains:

- `opaCluster` being deployable either as workloadKind `Daemonset` or `Deployment`
- `podDisruptionBudgets` only for workloadKind `Deployment`
- Affinities to fan out pods when `Deployment` is used across all nodes if possible
- Integration tests for swap `Daemonset` <> `Deployment`. Checks defaults for `localTrafficPolicy` is set correctly according to `workloadKind`
- Documentation on `workloadKind` mechanism, `PDBs` and default affinities.

### Author

- [x] CRD changes approved
- [x] CRD documentation for all fields, following the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).
- [x] Helm chart can be installed and deployed operator works
- [x] Integration tests passed (for non trivial changes)
- [x] Changes need to be "offline" compatible

### Reviewer

- [ ] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).
- [ ] Changelog updated

### Acceptance


Maleware:
I turned it back into a draft so it cannot be merged by accident. Still ready for review.

Maleware:
```
--- PASS: kuttl (1322.14s)
    --- PASS: kuttl/harness (0.00s)
        --- PASS: kuttl/harness/smoke_opa-1.16.2_openshift-false_use-tls-false (76.79s)
        --- PASS: kuttl/harness/openldap-user-info_opa-latest-1.16.2_openshift-false (88.01s)
        --- PASS: kuttl/harness/workload-kind_opa-latest-1.16.2_openshift-false (50.31s)
        --- PASS: kuttl/harness/smoke_opa-1.12.3_openshift-false_use-tls-true (79.86s)
        --- PASS: kuttl/harness/resources_opa-latest-1.16.2_openshift-false (19.21s)
        --- PASS: kuttl/harness/smoke_opa-1.12.3_openshift-false_use-tls-false (61.69s)
        --- PASS: kuttl/harness/logging_opa-1.16.2_openshift-false (79.47s)
        --- PASS: kuttl/harness/logging_opa-1.12.3_openshift-false (79.65s)
        --- PASS: kuttl/harness/aas-user-info_opa-latest-1.16.2_openshift-false (369.40s)
        --- PASS: kuttl/harness/config-overrides_opa-latest-1.16.2_openshift-false (58.94s)
        --- PASS: kuttl/harness/keycloak-user-info_opa-latest-1.16.2_keycloak-23.0.1_openshift-false (145.68s)
        --- PASS: kuttl/harness/smoke_opa-1.16.2_openshift-false_use-tls-true (66.54s)
        --- PASS: kuttl/harness/cluster-operation_opa-latest-1.16.2_openshift-false (54.29s)
        --- PASS: kuttl/harness/data-hub-resource-info_opa-latest-1.16.2_trino-latest-481_kafka-latest-4.2.1_superset-latest-6.1.0_data-hub-1.0.3_data-hub-prerequisites-0.3.0_openshift-false (1053.66s)
PASS
```

docs/modules/opa/pages/usage-guide/workload-kind.adoc
```diff
@@ -0,0 +1,84 @@
+= Workload kind
```
maltesander: This is missing from `nav.adoc`.

rust/operator-binary/src/crd/mod.rs
```diff
@@ -46,7 +46,7 @@ pub const DEFAULT_SERVER_GRACEFUL_SHUTDOWN_TIMEOUT: Duration = Duration::from_mi
 pub const SERVER_GRACEFUL_SHUTDOWN_SAFETY_OVERHEAD: Duration = Duration::from_secs(5);
 
 pub type OpaRoleType =
-    Role<OpaConfigFragment, OpaConfigOverrides, EmptyRoleConfig, GenericCommonConfig>;
+    Role<OpaConfigFragment, OpaConfigOverrides, v1alpha2::OpaRoleConfig, GenericCommonConfig>;
```
xeniape: More of a question and making sure: This change here would mean that the v1alpha1 version of the CRD also gets this v1alpha2 version field. This is intended? Maybe it's not fully clear in my head yet, but my assumption would be the point of having different CRD versions is to have different fields in there, but here we basically add something new to both versions. Since it only adds something and is therefore not a breaking change that's probably fine and is the reason why it was added in both versions?

rust/operator-binary/src/crd/mod.rs
```diff
@@ -143,6 +143,76 @@ pub mod versioned {
...
+        /// The Kubernetes workload the OPA servers run as.
+        ///
+        /// * `DaemonSet`: one Pod per node. `replicas` is ignored.
+        ///
+        /// * `Deployment`: fixed number of Pods, configured by `replicas`.
+        #[serde(default)]
+        pub workload_kind: WorkloadKind,
+
+        // `internalTrafficPolicy` is deliberately not a field here: the operator derives it from
+        // `workloadKind` in `OpaRoleConfig::internal_traffic_policy`. Exposing it as a user
```
xeniape: I don't understand the whole comment here but especially this line is confusing. It's saying `internalTrafficPolicy` is derived from this OpaRoleConfig struct here and there from the internal_traffic_policy field? Which does not exist?

rust/operator-binary/src/crd/mod.rs
```diff
@@ -143,6 +143,76 @@ pub mod versioned {
         pub tls: Option<OpaTls>,
     }
 
+    /// Role-level configuration for the OPA servers.
+    #[derive(Clone, Debug, Default, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct OpaRoleConfig {
+        /// The Kubernetes workload the OPA servers run as.
+        ///
+        /// * `DaemonSet`: one Pod per node. `replicas` is ignored.
```
labrenbe, line 152: When workload is DaemonSet should we log a warning that `replicas` is set but ignored?

rust/operator-binary/src/crd/mod.rs
```diff
@@ -143,6 +143,76 @@ pub mod versioned {
...
+    ///
+    /// Documentation:
+    /// [allowed Pod disruptions documentation](DOCS_BASE_URL_PLACEHOLDER/concepts/operations/pod_disruptions).
+    #[derive(Clone, Debug, Default, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct OpaPdbConfig {
+        /// Whether a PodDisruptionBudget should be written out for this role.
+        ///
+        /// Defaults to `true` when `workloadKind` is `Deployment` and to `false` when it is
+        /// `DaemonSet`, since a PodDisruptionBudget doesn't make sense for a DaemonSet.
```
labrenbe, line 207: Similarly should we log a warning that a PDB is configured but ignored when workload is DaemonSet?

tests/templates/kuttl/workload-kind/20-assert.yaml
```diff
@@ -0,0 +1,72 @@
+---
+# Switching `workloadKind` to `Deployment` swaps the workload object, widens the role Service's
+# traffic policy and adds a PodDisruptionBudget.
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 180
+commands:
+  - script: kubectl -n $NAMESPACE rollout status deployment test-opa-server-default --timeout 181s
+  - script: kubectl -n $NAMESPACE wait --for=condition=available opaclusters.opa.stackable.tech/test-opa --timeout 181s
```
labrenbe:
```suggestion
```
The replicas assertion in this file already test that the Deployment has been rolled out so we don't need these kubectl commands.

labrenbe:
```suggestion
```
The replicas assertion in this file already test that the Deployment has been rolled out so the kubectl commands only add noise to the kuttl output.

docs/modules/opa/pages/usage-guide/workload-kind.adoc
```diff
@@ -0,0 +1,84 @@
...
+        internalTrafficPolicy: Cluster # <1>
+  servers:
+    roleGroups:
+      default: {}
+----
+<1> Changes `internalTrafficPolicy` from Local to Cluster.
+
+== Changing the workload kind
+
+Changing `workloadKind` replaces the workload object, so policy queries can fail while the new Pods start up.
```
labrenbe, line 75: Maybe we could deploy both workloads in parallel and delete the old workload + change the service only once the new workload is available to avoid disruption.

docs/modules/opa/pages/usage-guide/operations/pod-placement.adoc
```diff
@@ -0,0 +1,33 @@
+= Pod placement
+:description: The default affinity spreads OPA Pods across Kubernetes nodes, which matters for a Deployment, and how to override it with your own affinities.
```
labrenbe, line 2: This sentence is a bit confusing to me

rust/operator-binary/src/crd/affinity.rs
```diff
@@ -0,0 +1,100 @@
...
+
+use crate::crd::{APP_NAME, OpaRole};
+
+/// Weight of the anti-affinity that spreads the Pods of a role across nodes.
+///
+/// The absolute value only matters once a second, competing term exists; see the `PreferSameNode`
+/// note on [`get_affinity`].
+const ANTI_AFFINITY_BETWEEN_ROLE_PODS_WEIGHT: i32 = 70;
+
+/// The default affinity of `role`: prefer to spread its Pods across nodes.
```
labrenbe, line 16: I have to say I find this doc comment a bit confusing. I'm not completely sure why, maybe because it's mixing a bunch of things...

rust/operator-binary/src/controller/build/resource/service.rs
```diff
@@ -229,6 +240,22 @@ mod tests {
         assert!(!spec.selector.unwrap().contains_key(ROLE_GROUP_LABEL));
     }
 
+    /// In `Deployment` mode the Pods do not cover every node, so node-local routing would strand
+    /// products running on Pod-less nodes. The policy has to follow `workloadKind`.
```
labrenbe, line 244:
```suggestion
    /// products running on nodes without OPA Pods. The policy has to follow `workloadKind`.
```

rust/operator-binary/src/controller/build/resource/service.rs
```diff
@@ -39,13 +39,24 @@ pub(crate) fn build_server_role_service(cluster: &ValidatedCluster) -> Service {
...
-        // unavailable, requests fail instead of falling back to another node.
-        // TODO: Once our minimum supported Kubernetes version is 1.35, use
+        // Derived from the role's `workloadKind`:
+        //
+        // * `Local` for a DaemonSet, so that products (e.g. Trino) on a node always talk to the OPA
+        //   Pod on the same node, avoiding cross-node latency. The downside is that if the local OPA
+        //   Pod is unavailable, requests fail instead of falling back to another node.
+        //
+        // * `Cluster` for a Deployment, whose Pods do not cover every node, so node-local routing
+        //   would leave products on Pod-less nodes unable to reach OPA at all.
```
labrenbe, line 49:
```suggestion
        //   would leave products on nodes without OPA Pods unable to reach OPA at all.
```

labrenbe (changes requested):
Looking good in general. My main concern is if switching workload kind really has to be disruptive.

rust/operator-binary/src/crd/mod.rs
Maleware: I wouldn't log it since in terms of a DaemonSet, replicas do not exist the way they would for a STS or Deployment. And since it cannot make sense I wouldn't be verbose about this.

rust/operator-binary/src/crd/mod.rs
Maleware: PDBs do not make sense in the eyes of a DaemonSet with `internalTrafficPolicy: Local` as any disruption of any pod would lead to impact on the platform components on the same node regardless how many pods are left on other nodes.

However, more generic, DaemonSets do not implement the `scale subresource` (according to [docs](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors)) means we cannot use `maxUnavailable`. Best we might be able to do is `minAvailable` which is an integer. Since a DS scales with node counts, `minAvailable` wouldn't give us the effect we'd want from a PDB.

Means, I wouldn't warn about ignoring something which was never intended to be used in the first place.

docs/modules/opa/pages/usage-guide/workload-kind.adoc
Maleware: I like the idea a lot and gave it some thought during implementation. I didn't do it for following reasons.

First, realising this would lead to a steep increase of resources used on a cluster. This of course depends on the replica and node count respectively, but can lead to a state where a `Deployment` or `DeamonSet` cannot be scheduled fully which leaves us somehow stuck half way in the switch.

Which leads directly to the second point: What is the exact definition of the "new" cluster being ready enough? Ports open? Pods being served? Status: ready? I wasn't able to come up with something decent to guarantee correct availability.

Last, building this would need at least some sort of a state machine. We would somehow need to persist a "mid-migration" marker (could be an annotation for example) which needs to be placed and removed according to the newly deployed clusters state, which is kinda complex and needs some code to orchestrate that.

All in all it was for me to much effort for a result which has a difficult to foresee and recover failure mode.

This said, if someone has a better idea how this can be achieved easily, I'd really like to have it.

rust/operator-binary/src/crd/affinity.rs
Maleware: To me it makes sense. Can you elaborate on what is confusing you?

docs/modules/opa/pages/usage-guide/operations/pod-placement.adoc
Maleware: Yeah :D Done with: [55dd7af](https://github.com/stackabletech/opa-operator/pull/873/commits/55dd7af195b3d254f3066d5f0b51ce0741ab36d0)

rust/operator-binary/src/crd/mod.rs
labrenbe: Similar to my comment on DaemonSet + PDB this warning would mainly be a hint for the user that they created a misconfig which will be ignored by the operator so that they can fix it.

rust/operator-binary/src/crd/mod.rs
labrenbe: I agree that deploying a PDB in this case would not make sense. 

So exactly because this is not intended to work I would warn the user that they made a misconfiguration which will be ignored and they don't wonder why there is no PDB even though they configured one.

rust/operator-binary/src/crd/affinity.rs
labrenbe: I'm not sure if the concerns belong to the TODO, the general affinity strategy or both. Also it's not clear to me what "role" is referring to in line 16? I assume it's the server role in the OPA spec?

